PR #614, opened 2021-11-04
feat(component-cookie-consent): react component cookie consent

nicolasBaudon:
## Description

This pr has the React implementation of the cookie consent. It is the same as the previous package but, we use `localStorage`, to reduce amount of cookies in the site, that means that we are not using any extra library to handle this.
I removed the prop `asuCookieDomain` as we cannot specify anything different, in `localStorage`, with that string.
I also added the complete documentation about the development and JSDOC Michele's implementation. 

### Component
image


mdibenedetto (commented):
Looks ok. 
I have only one major doubt.  I think any new package we add should follow the same folder structure we have used so far such:
src
-- components
-- - - $ComponentName
-- - - -- - - index.js
-- - - -- - - index.stories.js
-- - - -- - - index.test.js
-- core

I know we have only one component in this repo, but we never know the future development, 
and anyway have the same folder structure everywhere it may help when it is needed to run automation script.

mdibenedetto (approved):

mlsamuelson (changes requested):
Please fix errors .
```
lerna ERR! yarn run lint exited 1 in '@asu-design-system/components-core'
lerna ERR! yarn run lint stdout:
$ eslint --fix 'src/**/*.{js,jsx}' --ignore-path ../../.eslintignore

/Users/mlsamuel/web/uds-pr-review/packages/components-core/src/components/Video/index.js
  88:11  error  JSX props should not use functions  react/jsx-no-bind
  89:11  error  JSX props should not use functions  react/jsx-no-bind

✖ 2 problems (2 errors, 0 warnings)
```

mlsamuelson (changes requested):
@nicolasBaudon This looks great. Looks like there are some conflicts to resolve, then I'm happy to approve and merge this.

mlsamuelson (commented):
Approved.

PR #623, opened 2021-11-10
ASUIS 103 components

MarkLambe:
Just a draft PR to take a look at how these components are roughly architected. There's some tests, documentation, etc missing as I think the interfaces may change in the coming weeks.


mlsamuelson (changes requested):
Looking good except for the following error on build:
```
lerna ERR! yarn run lint exited 1 in '@asu-design-system/components-core'
lerna ERR! yarn run lint stdout:
$ eslint --fix 'src/**/*.{js,jsx}' --ignore-path ../../.eslintignore

/Users/mlsamuel/web/uds-pr-review/packages/components-core/src/components/TabbedPanels/index.test.js
  6:10  error  Button not found in '.'  import/named

✖ 1 problem (1 error, 0 warnings)

info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

mlsamuelson (approved):

PR #625, opened 2021-11-11
fix(bootstrap4-theme): stick footer to bottom of page

MarkLambe:
I'm not sure if we want to solve this problem this way but @duarte-daniela is having problems with the footer not sticking to the bottom of the page if there's less than 100vh content in total. This does resolve that.


mlsamuelson (changes requested):
Can we fix the merge conflicts before this is reviewed? Many thanks.

mlsamuelson (approved):

imorale2 (approved):

PR #626, opened 2021-11-11
fix: remove default max items and set to show all

nicolasBaudon:
## Description

Pr to remove default max items and when this field is not setted show all elements in feed.


mdibenedetto (approved):
LGTM

mlsamuelson (approved):

PR #647, opened 2021-11-26
fix: packages build optimization

nicolasBaudon:
## Description

In this pr I made the corresponding changes to remove the optimization warnings on packages build time. All of this changes were related to the media assets included in the build or the build output itself, so, I manage to update the packages to reduce bundle size, and update some media assets(depending of it usage) to not exceed the limit. 

For `app-rfi`, I wasn't possible to reduce it to the minimum but I lowered it from 1.02MiB to 347KiB that is 100KiB more that the recommended.

For `component-carousel`, the build was only 1KiB from the limit so I decide to extend that limit for that package, as it was insignificantly bigger.

The rest of the optimizations warning were resolved and remaining ones are about lint rules.

Ticket: https://asudev.jira.com/browse/UDS-1087?atlOrigin=eyJpIjoiYTNjNWM4NjU0MGQwNDE4ZWJiNmI0YzBjYWUxMmE3NGEiLCJwIjoiaiJ9


mlsamuelson (approved):
Love this. Thank you, @nicolasBaudon.

PR #678, opened 2022-02-01
fix(bootstrap4-theme): uds-content-align broken in some cases on mobile

MarkLambe:
Fixes [this ticket](https://asudev.jira.com/browse/ASUIS-181).


mlsamuelson (commented):
@MarkLambe Before I approve... This adjusts the scss for the `bootstrap4-theme` global header - do we need a similar adjustment for the `component-header` styles, as that's the header in use in the CMS?

mlsamuelson (approved):

PR #681, opened 2022-02-02
feat(app-webdir-ui): make various parts of profile, links

MarkLambe:
This PR is for [this](https://asudev.jira.com/browse/ASUIS-192) ticket.


mlsamuelson (changes requested):
This looks good. Including the use of the random photo. ;)

In reviewing I found a bit that looks like it could be addressed in this PR. The address data we have in the index isn't like street addresses we could feed into a url and generate a map from (that I'm aware of). They're university mailcodes or data similar to that. In the legacy site, addresses weren't clickable. For now, let's not make the address a clickable link here.

mlsamuelson (approved):

PR #682, opened 2022-02-07
As UI s 199 change opacity

MarkLambe:
This resolves [this issue](https://asudev.jira.com/browse/ASUIS-199). @mlsamuelson If you want me to do the fade over a fixed height let me know, though this happens in the small components (the 3-item list on the side) and the longer one (with 6 results) so it might have to take that into account.


mlsamuelson (approved):
@MarkLambe Looks good. No need to use a fixed height. We'll go with what they designers say. :)

PR #691, opened 2022-02-23
ASUIS-130 directory component

MarkLambe:
The UI for [this task](https://asudev.jira.com/browse/ASUIS-130).

Depending on what you want @mlsamuelson we can merge this now and then come back when decisions around the props and exact behaviour have been made, or we can wait. Whatever you think!


mlsamuelson (changes requested):
This threw quite a few errors during the build. When it won't break the build, I'd be happy to merge it in "stub" form. ;)

mlsamuelson (approved):

PR #803, opened 2022-07-08
fix(app-webdir-ui): added lock icon and changed components-core package

davidornelas11:


mlsamuelson (changes requested):
@davidornelas11 We need a commit for reach package with it in the commit scope in the conventional commit message so that they all get a release.

mlsamuelson (changes requested):
@davidornelas11 This mostly looks good.  I did notice there's a conflict that needs to be resolved - looks pretty minor. AND I'm seeing this on the Students tab - 2 sets of results messages:
Screen Shot 2022-08-30 at 2 27 16 PM

PR #820, opened 2022-07-28
fix(app-webdir-ui): authentication fix for click api

davidornelas11:
Ticket: https://asudev.jira.com/browse/ASUIS-751?atlOrigin=eyJpIjoiNGE0ZDAwOWE0NjM0NDVmZmJjZWY3ZGFlNmUwYTNhZDIiLCJwIjoiaiJ9


mlsamuelson (changes requested):

packages/app-webdir-ui/src/helpers/search.js
```diff
@@ -21,21 +21,33 @@ export const engineNames = {
...
-        reject(err);
-      });
+    let response;
+    if (props.loggedIn) {
+      const tokenResponse = await axios.get(`${props.API_URL}/session/token`);
+      const headers = {
+        "X-CSRF-Token": tokenResponse.data,
+      };
+      response = await axios.post(
+        `${props.API_URL}${props.searchApiVersion}webdir-click`,
```
scott-williams-az, line 38: Regarding building a url, at first glance these 2 lines look like an error
`${ props.API_URL }/session/token`
`${ props.API_URL }${ props.searchApiVersion }webdir-click`,

After I looked closer I was able to see the `searchApiVersion` contains a `/`
if we change searchApiVersion from `/api/v1/` to `/api/v1` then the following 2 lines look more consistent
`${ props.API_URL }/session/token`
`${ props.API_URL }${ props.searchApiVersion }/webdir-click`,

Just something to consider, no change required if we think this is not a huge concern. 
Could users technically add these props without proper `/`?
We just don't want the string to endup without any `/` joining the parts

davidornelas11: I originally tried changing that to what you proposed and ran into lots of errors because webspark users can use the webdir component and it is also used in the main isearch page. Unfortunately this was the only way to make it work

mlsamuelson: Yup, documented in https://github.com/ASU/asu-unity-stack/pull/820#discussion_r933697402

mlsamuelson (approved):

PR #837, opened 2022-08-31
Uds 1116 limit character

davidornelas11:


mlsamuelson (changes requested):
@davidornelas11 An unwanted `app-degree-pages` file update seems to have snuck in here. Let's remove it from the PR.

Also, I recall you indicating you wanted to discuss these changes in Post-Scrum? Let's do that after the unwanted file is removed.

PR #867, opened 2022-10-24
fix(bootstrap4-theme): fixed controls not showing on certain components

davidornelas11:


mlsamuelson (changes requested):
@davidornelas11 I noticed you rolled this out to some but not all `examples` and `stories`. I'm presuming those that had controls and were potentially affected.  Can we make the file naming consistent across all stories in the `bootstrap4-theme` so we avoid triggering this issue in the future if we add controls to a story or if someone is making a copy of a story but doesn't understand the implications of the naming?

mlsamuelson (changes requested):
I'm seeing path resolution errors when attempting to run `yarn storybook` in the `bootstrap4-theme` package. Can you resolve those?

mlsamuelson (changes requested):
@davidornelas11 This one is going to require some conflict resolutions after the build updates.

mlsamuelson:
@davidornelas11 New conflict to resolve on this one. 😭  Then let's finally get this one in! 🥲

mlsamuelson (changes requested):
One (hopefully) final conflict resolution.

mlsamuelson (approved):

PR #885, opened 2022-11-10
fix(app-webdir-ui): removed main headings on search page mobile results

davidornelas11:


maycomayco (commented):
LGTM

mlsamuelson (changes requested):
@davidornelas11 Looks like this only targets the All ASU results heading. We want to hide the local search results heading, too.

mlsamuelson (approved):
Nevermind. I misread the code in the PR. This looks good.

PR #886, opened 2022-11-10
feat(bootstrap4-theme): adapt hero mobile view to latest design

maycomayco:
These changes only affect to mobile version in the three variants SM, MD, and LG.

The change applies an empty div for styles purposes


maycomayco:
I used `div.hero-overlay` as a solution again

davidornelas11 (dismissed):

packages/bootstrap4-theme/src/scss/design-tokens/_variables.scss
```diff
@@ -320,6 +326,7 @@ $uds-component-heroes-uds-hero-uds-hero-container-h-1-margin-left: 0rem;
 $uds-component-heroes-uds-hero-uds-hero-container-uds-hero-text-margin-left: 0rem;
 $uds-component-heroes-uds-hero-uds-hero-container-p-color: #ffffff;
 $uds-component-heroes-uds-hero-uds-hero-container-p-font-size: 1.5rem;
+$uds-component-heroes-uds-hero-uds-hero-container-p-line-height: 1.75rem;
```
mlsamuelson: This variable was added to the `bootstrap4-theme`'s _variables.scss file which is a built file copied into place during the build. When trying to run storybook this has been overwritten and it crashes due to the missing `$uds-component-heroes-uds-hero-uds-hero-container-p-line-height`.

Instead, you'll want to update the design tokens so this variable gets created. The proper place to do this would be in https://github.com/ASU/asu-unity-stack/blob/dev/packages/design-tokens/components/heroes.json

Make the update, then run `yarn build` and check `packages/design-tokens/build/scss/_variables.scss` to locate the new variable that will have gotten created. The name of the variable is based on the location in the json structure for the new value.  This _variables.scss is copied into the `bootstrap4-theme` during build.

Let me know if you have questions.

mlsamuelson (changes requested):

mlsamuelson (approved):

PR #891, opened 2022-11-18
chore(component-header): set compatible range for React peer dependencies

coreyafunk:
Updates the React peer dependency version specifiers to allow for projects using React v16 latest or React v17.

This change only expands peer dependency support, so existing projects using this dependency should not be affected.

Same rationale as described in PR #806


mlsamuelson:
@coreyafunk I incorporated your change into that PR so there's now no changes in this. I'm going to close the PR.

coreyafunk:
> @coreyafunk I incorporated your change into that PR so there's now no changes in this. I'm going to close the PR.

Great, thank you!

PR #902, opened 2022-12-07
build: testing

mlsamuelson:


mlsamuelson:
Notes:
Lots of debugging required on this. But it's ready now.
Big hang up was needing to set the `runAsUser` to the node user, otherwise some internal user switching during the deploy scripts resulted in permissions issues in the file system.
Final commits cleanup the build. 
Misc notes:
* I renamed the stages to shorter names. 
* And because .npmrc and built files persist between stages, I was able to remove some unnecessary commands. 
* While the deploy stage completes successfully without doing another install and build, I kept that in as I believe it was there for ensuring interdependencies were up to date for the static site release, post publish.

Jenkinsfile
```diff
@@ -68,80 +76,59 @@ spec:
...
         }
-        stage('Deploy Storybook to GitHub Pages') {
+        stage('Deploy') {
             when {
-                branch 'dev'
+              branch 'dev'
             }
             steps {
                 container('node14') {
                     script {
```
maycomayco, line 110: Do we need to maintain this step with the scrip directive?

Jenkinsfile
```diff
@@ -68,80 +76,59 @@ spec:
...
+                        // TODO Remove after transition as it will be set in environment block:
+                        // Use Github token as NPM token with GH Packages
+                        NPM_TOKEN = GH_TOKEN
+
+                        // echo "Debug persisted .npmrc..."
+                        // sh 'cat ~/.npmrc'
+                        // sh 'yarn config list'
+
+                        echo '# Final, post-publish install and build...'
                         sh 'yarn install'
```
maycomayco, line 120: I understood that we don't need to run again the install and build commands. Is this ok?

Jenkinsfile
mlsamuelson: @maycomayco Yes, for now, in order to reassign the value of the NPM_TOKEN in this step on line 113. When we transition to the new registry fully, we can remove that and this no longer needs to be inside the script block.

package.json
```diff
@@ -22,8 +22,7 @@
     "percy": "lerna run percy",
     "start": "node server/server.js",
     "stop": "node server/server.stop.js",
-    "build-storybook": "storybook-to-aws-s3 --out=build --packages packages --bucket-path bucket/storybook --monorepo-index-generator server/storybook-index-generator.js --dry-run",
-    "deploy-storybook": "storybook-to-ghpages --out=build --packages packages --monorepo-index-generator server/storybook-index-generator.js --source-branch=develop-gha --ci",
+    "deploy-storybook": "storybook-to-ghpages --out=build --packages packages --monorepo-index-generator server/storybook-index-generator.js --source-branch=dev --ci",
```
davidornelas11, line 25: The deploy-storybook command runs the build-storybook internally, correct?

davidornelas11 (approved):

package.json
mlsamuelson: Yes. It takes care of all that. I removed the no longer needed build-storybook command to avoid confusion with the build-storybook command that Storybook itself surfaces.

PR #929, opened 2023-01-05
perf(bootstrap4-theme): update all incidences of class to className

maycomayco:
It's only for test purposes


mlsamuelson:
As per https://asudev.jira.com/browse/UDS-1196, we've decided NOT to proceed with this approach.

Thanks for all the hard work investigating this one, @maycomayco

PR #953, opened 2023-01-24
Remove react-reveal

davidornelas11:
Ticket: https://asudev.jira.com/browse/UDS-1238

### DIFF
--- a/packages/components-core/package.json
+++ b/packages/components-core/package.json
@@ -41,11 +41,12 @@
     "postdocs": "node ../../scripts/process-readme-props.js"
   },
   "devDependencies": {
-    "@percy/cli": "^1.0.0-beta.68",
-    "@percy/storybook": "^4.3.4",
     "@asu/bootstrap4-theme": "^1.0.0",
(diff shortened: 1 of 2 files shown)


packages/components-core/package.json
```diff
@@ -41,11 +41,12 @@
     "postdocs": "node ../../scripts/process-readme-props.js"
   },
   "devDependencies": {
-    "@percy/cli": "^1.0.0-beta.68",
-    "@percy/storybook": "^4.3.4",
```
mlsamuelson, line 45: Was the removal of the percy packages needed? They may not be needed ATM, but we'll be bringing Percy back before long. It's okay if they're causing issues.  Just hoping to avoid us spending half a day before realizing we regressed in the Percy settings when we go to bring it back online. ;)

davidornelas11: The dependencies got moved down further in the devDependencies section for some reason, but are still included

mlsamuelson (approved):
Percy bits were just re-sorted. :P 

Approved.

PR #970, opened 2023-02-14
feat(app-degree-pages): update layout for Required Courses section

maycomayco:
You will see many changes in the file, this is because the component file was getting very long and for this reason, I decided to make a small refactor always taking care of the readability of the code and performance.

**Outcome**
image

**Originally proposed design**
image

### DIFF
--- a/packages/app-degree-pages/src/components/DetailPage/components/RequiredCourse/index.js
+++ b/packages/app-degree-pages/src/components/DetailPage/components/RequiredCourse/index.js
@@ -1,21 +1,11 @@
 /* eslint-disable react/prop-types */
 import PropTypes from "prop-types";
 import React from "react";
-import Styled from "styled-components";
-
-import { Button } from "../../../../../../components-core/src/components/Button";
 
 /** @typedef {import('../../../../core/types/detail-page-types').RequiredCoursesProps} RequiredCoursesProps */
 
-const ButtonGroup = Styled.div`
-  display: flex;
-  flex-direction: column;
-  align-items: flex-start;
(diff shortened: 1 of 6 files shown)


packages/app-degree-pages/src/components/DetailPage/components/RequiredCourse/index.js
```diff
@@ -1,85 +1,119 @@
...
 /**
  *
  * @param {RequiredCoursesProps} props
  * @returns
  */
 function RequiredCourse({
-  concurrentDegreeMajorMaps = "",
   onlineMajorMapURL = "",
-  majorMapOnCampusArchiveURL,
+  majorMapOnCampusArchiveURL = "",
```
mlsamuelson: Love the improvements to readability, @maycomayco . I'd like to request a change - the `majorMapOnCampusArchiveURL` variable name should be `majorMapOnCampusURL` since it's not an archive link. At one point in the past that was a point of confusion, so let's remove it now.

mlsamuelson (changes requested):
For the default story, can we use the `acadPlan` `TBTGMBGM` ... here: https://github.com/ASU/asu-unity-stack/blob/UDS-1253/packages/app-degree-pages/src/components/DetailPage/index.stories.js#L96 since that will give us an example story that shows a major with subplans.

Also, while you're at that, the variable `queryAcadPllan` has a typo in the name (two l's) and should be fixed, too. ;) Not  your mistake, but good to get it fixed.

mlsamuelson (approved):
Thank you!

PR #973, opened 2023-02-17
fix(app-webdir-ui): update faculty rank tab 3 title, fix story defaul…

mlsamuelson:
…t template

[ASUIS-947](https://asudev.jira.com/browse/ASUIS-947)


davidornelas11 (approved):
Looks good!

PR #975, opened 2023-02-22
feat(bootstrap4-theme): image based cards component

maycomayco:
I added the new component and I also added support for dataLayer into window object

image


packages/bootstrap4-theme/src/scss/extends/_image-based-cards.scss
```diff
@@ -0,0 +1,146 @@
+.image-based-card {
```
mlsamuelson: When it's not a class name inherited by Bootstrap we have a convention of prefixing class names for Unity components with `uds-` so we avoid accidental collisions with other CMS/application classes and unexpected BS class names.  Can you change this to `.uds-image-based-card` and check for other class names that should be prefixed?

mlsamuelson (changes requested):

mlsamuelson (approved):
Thx for the addition of the prefix.

mlsamuelson:
This one still needs a Design Review by @imorale2 prior to merge.

mlsamuelson:
Met with @imorale2  in Zoom and did a design review. Approved.

PR #979, opened 2023-02-28
feat(bootstrap4-theme): add image based card component with CTA

maycomayco:
image
image
image


packages/bootstrap4-theme/src/scss/extends/_cards.scss
```diff
@@ -744,3 +745,158 @@ Cards - Table of Contents
...
+    z-index: 20;
+
+    .content-bg {
+      h3 {
+        font-size: $uds-size-font-xxl;
+        line-height: 2.125rem;
+        padding: $uds-size-spacing-3 $uds-size-spacing-3 $uds-size-spacing-1;
+        margin: 0;
+      }
+      .hidden-details {
```
davidornelas11: Should the hidden details be scrollable? Adding overflow: scroll to hidden-details fixes this but I'm not sure if that is the desired effect. Something to bring up in standup, maybe?

davidornelas11 (changes requested):

packages/bootstrap4-theme/src/scss/extends/_cards.scss
maycomayco: Good point to talk about in the standup today

davidornelas11 (dismissed):
Holding off on merging until Michael has some time to look it over as well

maycomayco:
I closed this one because it was covered here https://github.com/ASU/asu-unity-stack/pull/1070

PR #994, opened 2023-03-24
chore(bootstrap4-theme): remove dependency on design tokens (post BS5 package release)

scott-williams-az:
Design Tokens created many very specific variables for components which were used once
Kept common variables used multiple times

no change to output: packages/bootstrap4-theme/dist/css/bootstrap-asu.css 

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1278)


packages/bootstrap4-theme/src/scss/extends/_alerts.scss
```diff
@@ -22,12 +22,12 @@
     // Edits close button content (i.e. "X" inside bubble)
     .close
     {
-      opacity: $uds-component-button-close-opacity;
+      opacity: 1;
```
maycomayco, line 25: I see that many variables were changed, and it seems that this is not expected since those variables belong to the ASU design guidelines. Maybe it was a mistake, could you check it?

mlsamuelson: @maycomayco Are you saying that the calculated values have changed? What I see in the original design-tokens was that this specific value matches.

The overall approach matches my expectations for this task:
1. remove dependency on the design-tokens package
2. manage variables directly within the bootstrap package
3. avoid creating variables for single-use cases.

mlsamuelson (commented):
This looks good and tests out for me.  I've not marked as "approved" yet because I don't want it to get merged until after we have the final release for the Webspark 2.9.0 release compiled.

packages/bootstrap4-theme/src/scss/extends/_alerts.scss
maycomayco: LGTM!

mlsamuelson (changes requested):
Merge conflicts to resolve.

mlsamuelson:
Closing since we will be deprecating this component with the migration.

PR #995, opened 2023-03-24
chore(unity-bootstrap-theme): remove dependency on design tokens

scott-williams-az:
Design Tokens created many very specific variables for components which were used once
Kept common variables used multiple times

no change to output: packages/bootstrap4-theme/dist/css/bootstrap-asu.css 

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1278)


mlsamuelson (approved):
I didn't see any issues here. This can be merged into the `migration-to--bs5` branch when you're ready.

PR #997, opened 2023-03-28
feat(unity-bootstrap-theme): added html plugin

davidornelas11:


packages/unity-bootstrap-theme/.storybook/preview.js
```diff
@@ -9,4 +9,15 @@ export const parameters = {
...
     },
   },
+  html: {
+    prettier: {
+      tabWidth: 4,
+      htmlWhitespaceSensitivity: "ignore",
+    },
+    root: ".row", // can be customized to wrap an element
+    removeComments: /^\s*\s*$/,
+    transform: (code) => {
```
scott-williams-az, line 19: regex '.*' vs '.*?'
<svg.*<\/svg>: will make 1 match "<svg></svg><svg></svg>"
<svg.*?<\/svg>: will make match all "<svg></svg>", "<svg></svg>"
consider:
    transform: (code) => {
      return (code
        .replace(/<svg.*?<\/svg>/g, "")
        .replace(/()/gi, "$2"));
    }

packages/unity-bootstrap-theme/.storybook/preview.js
```diff
@@ -9,4 +9,15 @@ export const parameters = {
       date: /Date$/,
     },
   },
+  html: {
+    prettier: {
+      tabWidth: 4,
+      htmlWhitespaceSensitivity: "ignore",
+    },
+    root: ".row", // can be customized to wrap an element
```
scott-williams-az, line 17: I like the idea of adding a specific selector, like #html-root, this plugin will only match the first found selector so we should make it an id

scott-williams-az (changes requested):

packages/unity-bootstrap-theme/.storybook/preview.js
mlsamuelson: Agree.

PR #1016, opened 2023-04-19
build: update to node 18, move bootstrap4-theme to disabled

mlsamuelson:
### Description

Node 14 is EOL at EOM
Upgrade to Node 18 LTS. Disable `bootstrap4-theme` since it is to be deprecated and included complicating factors blocking the update to Node 18.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1333)

### Checklist

- DOES NOT build: Depends on- [UDS-1288](https://asudev.jira.com/browse/UDS-1288) effort being completed so React packages consume the new theme package
- SEE ABOVE: No new console errors


mlsamuelson:
Since this depends on UDS-1288 to build. Leaving this in DRAFT until that is in.

1) complete bootstrap5 package (possibly optional before next two steps)
2) Start and complete UDS-1288 to make package consume new theme
3) merge this PR

packages/app-degree-pages/package.json
```diff
@@ -97,7 +97,7 @@
     "webpack-merge": "^5.8.0"
   },
   "volta": {
```
scott-williams-az, line 99: Update Volta once in root package.json
All other updates can use `extends` 
"volta": {
    "extends": "../../package.json",
  }

https://docs.volta.sh/advanced/workspaces

mlsamuelson: Ohhh, nice.

mlsamuelson:
Removing BS4 resulted in the build breaking due to the location of the bootstrap package moving from within the package folder to the root node_modules (since there weren't 2 versions being installed now). 

I updated the webpack path for that and fixed broken include paths and in stories.

And lastly, chart.js was missing and throwing an error. I confirmed it had been included in the previous theme, so must have gotten lost with the removal of the old BS4 package.

packages/unity-bootstrap-theme/package.json
```diff
@@ -80,8 +81,7 @@
     "@fortawesome/fontawesome-free": "^5.15.3"
   },
   "volta": {
-    "node": "14.17.6",
-    "yarn": "1.19.1"
```
maycomayco, line 84: I ran into an issue when I tried to install dependencies in this package using `yarn 1.22.10`, for that I used `yarn 1.19.0`
We can leave it like this if everything works, but we should keep this in mind if we start having problems installing dependencies

mlsamuelson:
Closing this in favor of #1018

PR #1018, opened 2023-04-20
Uds 1333 node sass rm

davidornelas11:
### Description
Removed yarn.lock and jest-transform-css and jest-transform-scss  packages from component-carousel to fix node-gyp error `ERR! stack Error: `make` failed with exit code: 2` when trying to yarn install from root

### Checklist

- [x] No new console errors


.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
```
davidornelas11, line 5: Left this in on accident but it is my config for debugging through vscode instead of the browser.

mlsamuelson: Sneaking a developer experience improvement into the platform with this commit... but it's so well-documented, and ancillary. What's one more file in a PR with 1.2k+ ? 😂 

We can pull it out later if there are issues, but shouldn't get in the way of anything.

mlsamuelson:
I've created https://asudev.jira.com/browse/UDS-1339 to address the need to update the testing for `composer-carousel` in light of this solution.

PR #1033, opened 2023-05-08
fix(app-webdir-ui): fix sideways profiles by adding parameters

tbutterf:
### Description

There is an issue where uploaded profile images are sometimes displayed sideways.

The CMS solves this problem by adding parameters to the URL when requesting the image from the photo web service. The parameters are `?size=medium&break=[num]`. The `size=medium` parameter apparently forces the image to be displayed in profile orientation, rather than landscape. The `break=[num]` parameter breaks the cache on the server and forces a new image to be loaded.

Since this seems to work great on the CMS side, I figured we needed to do a similar treatment to images being displayed in the component. This PR brings the component into alignment with the CMS.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1200)

### Checklist

- [N/A] Add/updated examples
- [x] No new console errors
- [N/A] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### Images
CMS version
image
Updated component
image


davidornelas11 (dismissed):
to test, department ID 192466 can be used for department type web directory

packages/app-webdir-ui/src/helpers/dataConverter.js
```diff
@@ -239,7 +241,7 @@ export const staffConverter = ({
           id={asuriteEID}
           profileURL={`${profileURLBase}/profile/${asuriteEID}`}
           key={asuriteEID}
-          imgURL={filledDatum.photo_url.raw}
+          imgURL={`${filledDatum.photo_url.raw}${queryConcat}size=medium&break=${dateVal}`}
```
maycomayco: Could we avoid formatting the url in the prop? I don't want to sound picky, but I think it would be an improvement in readability if we create a constant with the absolute value of the URL.

tbutterf: Okay. I have pushed up a change to account for this.

maycomayco (dismissed):
LGTM, nice job!

davidornelas11 (dismissed):
Looks good!

mlsamuelson (changes requested):
With the way React re-renders pages, the use of the `break=NNN` value, being attached to the current timestamp will mean that the image gets re-requested at every re-render... e.g. if I page forward and back, even though the page hasn't reloaded, that'll be a new request.  What about guarding against hammering the photos service by setting the last one or two digits of the timestamp to 0 - effectively giving a 10sec or 100sec window where the browser can choose to not re-load the image?

mlsamuelson:
@maycomayco Looks like the "deadzone" on the timestamp is too aggressive, see my comment on https://asudev.jira.com/browse/UDS-1200?focusedCommentId=1767553

mlsamuelson (changes requested):
This looks great. One last task is to also apply `format_imageUrl()` for the `imgURL` parameter in the `studentsConverter()` function ... see line 320 of `packages/app-webdir-ui/src/helpers/dataConverter.js`. This will ensure we apply the same photo logic for students like we currently are for Faculty and Staff. To confirm, you can look at students results in the Search Page story in Storybook.

maycomayco:
I added a TODO comment for the future, you can see it here https://github.com/ASU/asu-unity-stack/pull/1033/commits/9a074b0a2c4738fcd6b9c478cb7a737f56773f4b#diff-4e42e24adf181bae5325a04b81fc59601bebb05ddce4908e06fd66e081d44c36R48
cc @mlsamuelson

mlsamuelson:
@maycomayco Yes, that's been a painful issue on the implementation side. We have a request in with the Photos WebService team so that we can add a parameter to the image url that will tell it to return a Web Standards-compatible default image when there is none, instead of a 404. So that will resolve that issue. :)

mlsamuelson (approved):
Looks good! Thank you, @maycomayco !

PR #1039, opened 2023-05-16
chore(unity-bootstrap-theme): dataLayer example is readable

scott-williams-az:
### Description
Removed shorthand code and made data layer example more clear for others to implement

### Links

- codepen example of how the datalayer was incorrect https://codepen.io/stwill/pen/QWZVWVe

### Checklist

- [x] Add/updated examples


gulpfile.js
```diff
@@ -9,6 +9,7 @@ copy = () => gulp.src(["./server/views/**/*"]).pipe(gulp.dest("./build/@asu"));
 // in the /server folder.
 // https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/troubleshooting-custom-domains-and-github-pages#cname-errors
 cname = () => gulp.src(["./server/CNAME"]).pipe(gulp.dest("./build/"));
+cname = () => gulp.src(["./server/dataLayerExample.js"]).pipe(gulp.dest("./build/"));
```
davidornelas11: No errors are thrown here but this should probably have another name. It seems to overwrite the previous cname function

mlsamuelson: Yes that concerns me, too. Maybe name it `datalayer`. And then you'll have to include it in the series at the bottom of the file.

mlsamuelson (changes requested):

gulpfile.js
mlsamuelson: Looks like that had been an interim state. Marked it resolved.

maycomayco (approved):
LGTM, great job Scott! 👏
That codepen should be included as a live example in our documentation!

PR #1049, opened 2023-05-18
fix(asu-unity-stack): sitename position in header

maycomayco:
### Description

We had an error in HTML markup, I fixed that in all views of the monorepo index.
Also added a friendly output when you run `yarn start` command

There still seems to be a subtle difference between the index page and the others, which might be caused by some variation in our HTML. In the next steps, we should make some improvements to our templates to avoid code duplication, among other things.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1368)

### Checklist

- [x] Add/updated examples
- [x] No new console errors

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### Images

image

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -1,7 +1,7 @@
-const express = require("express");
-const path = require("path");
-const nunjucks = require("nunjucks");
 const compression = require("compression");
+const express = require("express");
+const nunjucks = require("nunjucks");
+const path = require("path");
 
 const app = express();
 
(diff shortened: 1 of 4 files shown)


mlsamuelson:
@maycomayco Great thoughts on future enhancements. We have plans to redo the static site in the near future. We could make this a lot simpler/better. It was originally designed to run in a NodeJS express server which we don't do anymore with the GHPages approach.

mlsamuelson (approved):

PR #1065, opened 2023-05-23
fix(app-webdir-ui): rework homepage default tab content

tbutterf:
UDS-1393

### Description

The search home page has three tabs with exactly the same content on each tab. This is confusing and bad user experience.

To solve this issue, I have created different content for each of the tabs. This is my first attempt at doing something like this in React, so, if I have done something incorrectly, please let me know and I will change it.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1393)

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### Images

image
image
image


packages/app-webdir-ui/src/SearchPage/index.js
```diff
@@ -98,10 +100,16 @@ function SearchPage({
...
+  const componentMap = {
+    FacultyTab: <PreFacStaffMsg />,
+    StudentsTab: <PreStudentMsg />,
+  };
+
   const preSearchOrContent = content => {
     if (term) return content;
 
-    return <PreSearchMessage />;
+    const component = componentMap[content.type.name];
```
davidornelas11: Just tested this in an isearch ddev environment and react does not have the content.type.name info so it always comes up with the initial search page info. I think this might require using hooks to render that message. Might be a bit more complex than this unfortunately

davidornelas11 (changes requested):

packages/app-webdir-ui/src/SearchPage/index.js
tbutterf: Weird. I'll look into this more...

tbutterf: Thanks @davidornelas11 for looking at this. I have no idea how to fix it. Haha. Maybe I need to hand this ticket off to you or Scott or Mayco. I thought it would be a fairly simple fix, but I guess it's more complicated. Dang. Oh well. It was worth a shot. 😄

mlsamuelson:
I have some thoughts on wording tweaks.

Let's not say "Faculty/Staff" and instead use "Faculty and Staff" like on the tab.

For students:
Screenshot 2023-05-23 at 2 29 34 PM

`
Students must opt-in to make their profile publicly available. The remainder of student profiles that do not have directory holds on them are available to ASU affiliates with an ASURITE ID and can be be viewed by logging into the site.
`
Consolidates all the text into one paragraph.
Changes "Id" to "ID".
Simplify language about logging in - make the link be "logging into the site"

tbutterf:
I have incorporated @davidornelas11's fix, merged from dev, and made the changes @mlsamuelson suggested. I think this is good to go now.

davidornelas11 (dismissed):
Looks good and tested in an isearch environment!

mlsamuelson (dismissed):
We're kinda missing out on the best way to leverage React components here. Let's do away with the `PreStudentMsg` and `PreFacStaffMsg` components and handle those as variations in the existing `PreSearchMsg` component. You can add a prop that gets passed into that or maybe get the current `searchTabsId` and use that to determine the content to display. This will ensure we make the best use of components as "templates" and will guard against unnecessary proliferation of components.

mlsamuelson:
Tabbed Panels are getting a bit of a rework in https://asudev.jira.com/browse/UDS-1412, so I'm going to put this into Draft state, as discussed with @tbutterf

mlsamuelson (approved):

PR #1087, opened 2023-06-08
Uds 1411 - Cards need to have links in titles

davidornelas11:
### Description
Cards need to have titles that are links

Titles on cards for events and news now have links in the titles

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1411?atlOrigin=eyJpIjoiY2QyNjA5ZmVlMzNmNGVjNTg0Yzc5ZDhhMThlYjllY2YiLCJwIjoiaiJ9))

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### DIFF
--- a/packages/component-carousel/src/components/CardCarousel/index.js
+++ b/packages/component-carousel/src/components/CardCarousel/index.js
@@ -80,7 +80,6 @@ const CardCarousel = ({
   const carouselItems = cardItems.map(item =>
     htmlTemplate(item, cardType, cardHorizontal, cardEventFormat)
   );
-  console.log(cardItems)
   const activateGlideActions = cardItems.length > perView;
 
   return (
--- a/packages/component-carousel/src/components/CardCarousel/index.js
+++ b/packages/component-carousel/src/components/CardCarousel/index.js
@@ -57,7 +57,7 @@ const htmlTemplate = (
       eventTime={eventTime}
       buttons={buttons}
       linkLabel={linkLabel}
-      linkUrl={linkUrl}
+      linkUrl={linkUrl || buttons?.[0]?.href}
       tags={tags}
     />
   ),
@@ -80,6 +80,7 @@ const CardCarousel = ({
   const carouselItems = cardItems.map(item =>
     htmlTemplate(item, cardType, cardHorizontal, cardEventFormat)
   );
+  console.log(cardItems)
   const activateGlideActions = cardItems.length > perView;
 
   return (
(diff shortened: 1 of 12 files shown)


mlsamuelson (commented):
For `component-events` we need to be using the `alias` field for the event title's link. We can't count on the rsvp links to be there.

packages/component-carousel/src/components/CardCarousel/index.js
```diff
@@ -57,7 +57,7 @@ const htmlTemplate = (
       eventTime={eventTime}
       buttons={buttons}
       linkLabel={linkLabel}
-      linkUrl={linkUrl}
+      linkUrl={linkUrl || buttons?.[0]?.href}
```
mlsamuelson, line 60: @davidornelas11 I debated on whether this update belonged in here, but after testing, seeing that it brings the behavior in-line for the header link regarding the "read more" `linkUrl` and the `buttons`, I think ultimately it makes sense. Good call.

packages/components-core/src/components/Card/index.js
```diff
@@ -316,9 +316,10 @@ const CardContent = ({
   tags,
 }) => (
   <>
+  {console.log(linkUrl, linkLabel)}
```
mlsamuelson: `console.log()`

mlsamuelson (changes requested):

mlsamuelson (changes requested):
@davidornelas11 Still needs work. We should be using using the `alias` field for the event title's link in `component-events`. We can't count on the rsvp links to be there - instead we should be linking to the `alias` which is the url for the event, so it'll always be there.

packages/component-events/src/core/transformers/data.transfromer.js
```diff
@@ -12,7 +12,7 @@ const transformData = data => ({
   interests: data.node.interests,
   startDate: data.node.very_start_date,
   endDate: data.node.very_end_date,
-  alias: data.node.alias,
+  alias: data.node.alias, // Also used for card link in title of card
```
davidornelas11, line 15: Kept this the same, only added a comment for clarity on how we use it

packages/component-news/src/components/CardGridNews/index.js
```diff
@@ -24,13 +24,14 @@ const gridRow = (feed, cardButton) => (
       eventFormat="inline"
       eventLocation={feed.location}
       clickable={!!feed.buttonLink}
+      cardLink={feed.alias}
```
davidornelas11, line 27: matches what we do in component-events, but uses alias without the date, which does not seem to be available for news data

scott-williams-az (approved):

mlsamuelson (approved):

PR #1102, opened 2023-06-28
feat(components-core): new Image component

maycomayco:
### Description

New image component for UDS

Some thoughts on these changes: 
- The configuration of the `.prettierrc` file was reverted as it was unnecessarily changing the files in the entire project. I don't consider it necessary to make this change now.
- I had to use the `// eslint-disable-next-line import/no-cycle` rule because the `utils.js` file has different responsibilities. On one hand, it imports all the utils classes like the `spreadClasses()` method, and it also exports the `init*()` methods so that the components can be used outside of UDS. This will be something to review in the future for better code modularization.
- Also, some props defined in the Testimonial component stories were misdefined.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1358)

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### Images

image

### DIFF
--- a/packages/components-core/src/components/RankingCard/index.js
+++ b/packages/components-core/src/components/RankingCard/index.js
@@ -1,27 +1,28 @@
-import classNames from 'classnames';
-import PropTypes from 'prop-types';
-import React, { useState } from 'react';
+import classNames from "classnames";
+import PropTypes from "prop-types";
+import React, { useState } from "react";
 
-import { trackGAEvent } from '../../core/services/googleAnalytics';
-import { sanitizeDangerousMarkup } from '../../core/utils/html-utils';
-import { Image } from '../Image';
+import { trackGAEvent } from "../../core/services/googleAnalytics";
+import { sanitizeDangerousMarkup } from "../../core/utils/html-utils";
+// eslint-disable-next-line import/no-cycle
+import { Image } from "../Image";
 
 const gaDefaultObject = {
-  name: 'onclick',
-  event: 'link',
-  action: 'click',
-  type: 'internal link',
-  region: 'main content',
+  name: "onclick",
+  event: "link",
+  action: "click",
+  type: "internal link",
+  region: "main content",
 };
 
 const AVAILABLE_GA_ACTIONS = {
-  OPEN: 'open',
-  CLOSE: 'close',
+  OPEN: "open",
+  CLOSE: "close",
 };
 
 const AVAILABLE_SIZES = {
-  LARGE: 'large',
-  SMALL: 'small',
+  LARGE: "large",
+  SMALL: "small",
 };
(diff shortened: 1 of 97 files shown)


packages/components-core/src/components/RankingCard/index.js
```diff
@@ -1,56 +1,44 @@
...
     <div className="image-wrapper">
-      <img
-        src={image}
-        alt={imageAlt}
-        loading="lazy"
-        decoding="async"
-        // eslint-disable-next-line react/no-unknown-property
-        fetchpriority="high"
-      />
+      <Image url={image} alt={imageAlt} fetchPriority="high" />
```
mlsamuelson: Image doesn't render on Ranking Cards. I believe it's because `url` here should actually be `src`.

packages/components-core/src/components/RankingCard/index.js
```diff
@@ -1,56 +1,44 @@
...
   ) : (
-    <img
-      src={image}
-      alt={imageAlt}
-      loading="lazy"
-      decoding="async"
-      // eslint-disable-next-line react/no-unknown-property
-      fetchpriority="high"
-    />
+    <Image url={image} alt={imageAlt} fetchPriority="high" />
```
mlsamuelson: Image doesn't render on Ranking Cards. I believe it's because `url` here should actually be `src`.

mlsamuelson (changes requested):
This is looking good. Found an issue with Image in the ranking cards (see comments attached to code).  

Some things I'd like to discuss with the team as a whole:
* handling of the .prettierrc.js `singleQuote` setting. Perhaps we should have a PR (or one per each package may be better, for size) before this that peforms ONLY cleanup to make quotes consistent. Then this PR could be updated so we don't have to regress that.
* Also, the eslint disable `import/no-cycle` ... since we don't need to rush into this change, I'd like to see us spend more time to identify the optimal way to handle that without using a disable rule before bringing this into the platform, to help us avoid unintended consequences.

mlsamuelson:
We agreed on reverting the prettierrc.js quote setting via Slack.

Conflicts to resolve here now.

mlsamuelson (approved):

mlsamuelson (approved):
This looks good to me. I'll create a ticket for resolving the `import/no-cycle` in the future.

Of note: the `packages/components-core/src/core/utils/index.js` doesn't provide an init function for usage of the Image component in UMD/browser-based use cases, but that makes sense - in most of those cases there will be another app managing images, so the usage of the Image component is more directly for within other UDS packages.

PR #1104, opened 2023-07-06
style: inactive colors for vscode

mlsamuelson:
### Description

Update to help with context-switching. Background: UDS VSCode titleBar styles disappear when the window is inactive, unless defined like here. I've chosen to set the foreground text to a UDS-compatible gray when inactive so we retain a cue to inactivity while keeping the background of the titleBar a consistent maroon so the window is easily recognizable as a UDS window.


davidornelas11 (approved):
LGTM

PR #1113, opened 2023-07-18
fix(components-core): align anchor menu with its content

scott-williams-az:
### Description

Issue of the problem found on app-degree-pages, which uses components-core anchor menu.
The anchor menu hard codes the container-xl

### Solution

The component will look at the content of the first item and align with those containers found or will default to container-xl
Also added some performance to avoid re-rendering:
- optimized functions with multi state updates to only update state once
- moved sticky to state, bad practice to use a ref when we can use a state
- added throttle and debounce functions to limit scroll events

Alternate solution would be to add a prop for container

### Links

- [Unity reference site](https://unity.web.asu.edu/@asu/components-core/iframe.html?args=&id=uds-anchormenu--default&viewMode=story)
- [JIRA ticket](https://asudev.jira.com/browse/UDS-1435)
- [Unity Design Kit](https://xd.adobe.com/view/56f6cb78-9af5-4b12-b4ce-ef319f71113f-03a5/screen/8f348a6b-fb44-439d-b739-07ec4886101d/)

### Checklist

- [ ] Add/updated examples
- [ ] No new console errors
- [ ] Accessibility checks

### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### DIFF
--- a/packages/components-core/src/components/AnchorMenu/index.js
+++ b/packages/components-core/src/components/AnchorMenu/index.js
@@ -36,10 +36,11 @@ export const AnchorMenu = ({
   const isSmallDevice = useMediaQuery("(max-width: 991px)");
   const [state, setState] = useState({
     hasHeader: false,
-    hasDegreeContainer: false,
+    hasAltMenuSpacing: false,
     containerClass: "container-xl",
     activeContainer: "",
     showMenu: false,
+    sticky: false,
   });
   const headerHeight = isSmallDevice ? 110 : 142;
 
@@ -88,8 +89,8 @@ export const AnchorMenu = ({
     debounce(handleWindowScroll, timeout);
   };
 
-  // Set any ASU Header on the document
-  const getHasHeader = () => {
+  // Is ASU Header on the document
+  const isHeader = () => {
     const pageHeader =
       document.getElementById("asu-header") ||
       document.getElementById("headerContainer") ||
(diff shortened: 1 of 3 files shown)


packages/components-core/src/components/AnchorMenu/index.js
```diff
@@ -66,44 +70,86 @@ export const AnchorMenu = ({
...
       document.getElementById("headerContainer") ||
       document.getElementById("asuHeader");
-    setHasHeader(!!pageHeader);
+    return (!!pageHeader);
   };
 
   // Sets prop for styled-component to change anchor menu style
   // based on if it requires different spacing for the ASU Header
-  const findContainersRequiringAltMenuSpacing = () => {
+  const getHasDegreeContainer = () => {
```
mlsamuelson: @scott-williams-az When originally developed, I asked @davidornelas11 to change this function name to something more general, since it's possible we may discover other use cases and having a general name means we're more likely to see the opportunity to leverage it in the future rather than accidentally developing a parallel solution. Original request at https://github.com/ASU/asu-unity-stack/pull/847#pullrequestreview-1140926753

Thoughts, Scott, Dave? I'm inclined to keep it a general name as it then "feels" more loosely coupled and extendable, conceptually.

mlsamuelson (changes requested):
Some great work here. Thank you for considering the performance implications and adding the throttle and debounce utils.

packages/components-core/src/components/AnchorMenu/index.js
scott-williams-az: This function is very specifically looking for an element with ID `degreeDetailPageContainer` so a specific function name I think is OK.
However...
At least now there is only one call to this function so we probably do not need a function at all.
Just do the logic in place at line 135
~hasDegreeContainer: getHasDegreeContainer(),~
`hasDegreeContainer: (!!document.getElementById("degreeDetailPageContainer")),`

mlsamuelson: I think what you suggest makes sense in handling it as a one-off in one line. If we find ourselves repeating this in the future, then would be the time to build out something more extendable.

mlsamuelson (approved):
Looks good. Included a commit to trigger a release for `app-degree-pages` as well, to correspond.

PR #1133, opened 2023-08-01
WS2-1634: fix(app-degree-pages): add check to the anchorMenu for false values

bebaps:
### Description

Currently, the Anchor Menu on the degree pages renders by default with no items. In this state, there is a bug where if you scroll down the page, then scroll back up the page, the menu will then disappear completely.

This PR adds a check to the `OnThisPageAnchorMenu` component to see if all values from the `anchorMenu` prop are false. If true, the component returns `null`.

This really may not be the best place to put this logic, or returning null for a component might just be a terrible idea. The end goal is to prevent the component from rendering at all if all the initial `anchorMenu` values are false. Please suggest changes as needed.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/WS2-1634)

### Checklist

- [x] No new console errors

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox

### DIFF
--- a/packages/app-degree-pages/src/core/components/OnThisPageAnchorMenu/index.js
+++ b/packages/app-degree-pages/src/core/components/OnThisPageAnchorMenu/index.js
@@ -15,11 +15,6 @@ import { anchorMenuPropType, progDetailSectionIds } from "../../models";
  * @returns {JSX.Element}
  */
 function OnThisPageAnchorMenu({ anchorMenu }) {
-  // Check if all keys in anchorMenu are false
-  if (Object.values(anchorMenu).every(value => value === false)) {
-    return null;
-  }
-
   // clean up those targetIdNames bad formatted
   /** @type {AnchorMenuItem[]} */
   const externalAnchors = anchorMenu?.externalAnchors?.map(anchorItem => {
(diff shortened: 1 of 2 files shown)


scott-williams-az (commented):
Nice job, it's always good to check values before rendering, I noticed we are also filtering out the false values below line 38,

React has a shortcut which will check a condition before rendering...

the format looks like this 
`return (condition && <Component/>)`

in order to keep the filter logic in one place can we update to the following?

```
return (
    anchorList.length > 0 && (
      <AnchorMenu
        items={anchorList}
        firstElementId={anchorList[0]?.targetIdName}
        focusFirstFocusableElement={false}
      />
    )
  );
```

scott-williams-az (changes requested):
previous comment

scott-williams-az (changes requested):
This logic might be better in components-core https://github.com/ASU/asu-unity-stack/blob/dev/packages/components-core/src/components/AnchorMenu/index.js#L183
```
  return (
    items?.length > 0 && (
      <AnchorMenuWrapper...
      </AnchorMenuWrapper>
    )
  );
```

bebaps:
Comment to myself so I don't forget, update this PR as follows:

- implement Scott's most recent suggestion
- remove my code
- verify all working as expected

bebaps:
@scott-williams-az requested changes made, verified they work as expected.

scott-williams-az (approved):
LGTM

mlsamuelson (approved):

PR #1143, opened 2023-08-10
feat(app-webdir-ui): added alpha sort on web directory

davidornelas11:
### Description
Alpha sort on web directory to allow for last initial sorting

When testing, you must change webdirectory story props to an api endpoint with the new alpha sort feature available. Also adding alphaFilter in the props as "true" will enable it

### Links

- [Unity reference site](https://unity.web.asu.edu/@asu/app-webdir-ui/index.html?path=/story/atoms-presearchmessage-template--pre-search-message)
- [JIRA ticket](https://asudev.jira.com/browse/SCHWEB-627?atlOrigin=eyJpIjoiOGIwNjYzNjI2ZWYyNDViMmI1ZTcwODkwYTI5YmRjODEiLCJwIjoiaiJ9)

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### DIFF
--- a/packages/app-webdir-ui/src/FacultyRankComponent/index.js
+++ b/packages/app-webdir-ui/src/FacultyRankComponent/index.js
@@ -1,11 +1,18 @@
 import { TabbedPanels, Tab } from "@asu/components-core";
 import PropTypes from "prop-types";
-import React from "react";
+import React, { useState, useEffect } from "react";
 
+import { FilterComponent } from "../helpers/Filter";
 import { engineNames, engines } from "../helpers/search";
 import { ASUSearchResultsList } from "../SearchResultsList";
 
-export const FacultyRankTabPanels = ({
+const filtersData = {
+  1: "Faculty",
+  2: "Academic Professionals",
+  3: "Fixed-Term Faculty and Academic Professionals",
+};
+
+const FacultyRankTabPanels = ({
   filters,
   API_URL,
   searchApiVersion,
@@ -14,9 +21,19 @@ export const FacultyRankTabPanels = ({
   display,
   profileURLBase,
   searchType,
+  alphaFilter,
 }) => {
+  const [requestFilters, setRequestFilters] = useState(null);
+  const [tabChange, setTabChange] = useState(null);
   const RES_PER_PAGE = 6;
   const sort = "faculty_rank";
+
+  useEffect(() => {
+    const tempFilters = { ...filters };
+    tempFilters["deptIds"] = deptIds.split(",");
+    setRequestFilters(tempFilters);
+  }, [deptIds, tabChange]);
+
   const engineParams = {
     filters,
     API_URL,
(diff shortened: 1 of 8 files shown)


mlsamuelson (changes requested):
This is looking nice. I like it. A few things I found:
* Let's add the `alphaFilter` prop to at least one of the example stories so it's easy to confirm. Bonus points if it could be added as a control in Storybook, but if that's a lot of work, it's not necessary for this component.
* `alphaFilter` prop - this is defined as a string, but really operates as a boolean. Is there a reason behind this decision, or should we consider converting to boolean? I'll grant that sometimes with how we convert and transmit the props through various systems, a string might be less prone to data-type issues. 
* People and People in Department stories don't seem to be honoring the alpha filters when I select them.

mlsamuelson:
@davidornelas11 I suspect this may be the issue with the two stories that don't honor the filter: https://github.com/ASUWebPlatforms/asu-isearch/pull/503#pullrequestreview-1590066411 Lack of support for the endpoint used in these cases. TBD

davidornelas11:
> 

The reason for the string choice is because in the past we have had problems with boolean props coming through as strings and needing to be converted. If the prop will be coming in as a boolean then i can definitely change it

mlsamuelson:
@davidornelas11 Did a design review with @imorale2 . She's thinking through the scrollbar... more to come on that. Discussion on how we were able to remove the scrollbar from Tabbed Panels may be instructive here: https://asudev.jira.com/browse/UDS-1355 but the constraints are different. Worth knowing, though.

Some feedback to address while waiting on further thoughts:
1. We need the same whitespace below the alpha filter as we have above it.
2. Let's make the font bold
3. Focus on the filter shouldn't be the blue glow - it should be the black outline.
4. While tabbing over the component, focus skips over the alpha filter, and it needs to be focusable without a click.

davidornelas11:
> @davidornelas11 Did a design review with @imorale2 . She's thinking through the scrollbar... more to come on that. Discussion on how we were able to remove the scrollbar from Tabbed Panels may be instructive here: https://asudev.jira.com/browse/UDS-1355 but the constraints are different. Worth knowing, though.
> 
> Some feedback to address while waiting on further thoughts:
> 
> 1. We need the same whitespace below the alpha filter as we have above it.
> 2. Let's make the font bold
> 3. Focus on the filter shouldn't be the blue glow - it should be the black outline.
> 4. While tabbing over the component, focus skips over the alpha filter, and it needs to be focusable without a click.

I added the formatting for the spacing and also added the same arrow as in tabbed panels. For accessibility, a user has to use the arrows to scroll through since if they needed to tab through, they would essentially be stuck in there until getting through Z

tbutterf:
@davidornelas11 and @mlsamuelson Because the `/api/v1/webdir-profiles/department` API endpoint is a POST-only endpoint, it does not use the filters that are set up for the other endpoint. So, we needed to add the `last_init` parameter to the `axios.post` call. This enables the back-end changes I have made to work properly. So, I have added this in a new commit.

mlsamuelson:
We got issues!

@davidornelas11 Build errors:
```
       /Users/mlsamuel/web/asu-unity-stack/packages/app-webdir-ui/src/helpers/Filter/index.js
         90:13  error  'trackArrowsEvent' is not defined  no-undef
         94:13  error  'trackArrowsEvent' is not defined  no-undef
```

@tbutterf I don't understand the issue you surfaced and the solution. Is this change backwards compatible for others?

@imorale2 Did you get back to Dave on the design concerns?

mlsamuelson (approved):

PR #1145, opened 2023-08-15
fix(components-core): improve pagination accessibility, remove unused…

davidornelas11:
… props

Added accessibility in pagination component and removed unused props

BREAKING CHANGE: Removed unused props

### Description
Fixed accessibility in pagination component and remove unused props

### Links

- [Unity reference site](https://unity.web.asu.edu/@asu/components-core/index.html?path=/story/uds-pagination--default)
- [JIRA ticket](https://asudev.jira.com/browse/UDS-1311?atlOrigin=eyJpIjoiNGIzZTg5Y2JhYzkwNDkyMjhhOTM1OTkxMGM2MmRhYjQiLCJwIjoiaiJ9)

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### DIFF
--- a/packages/components-core/src/components/AnchorMenu/index.js
+++ b/packages/components-core/src/components/AnchorMenu/index.js
@@ -227,7 +227,11 @@ export const AnchorMenu = ({
               [`show`]: state.showMenu,
             })}
           >
-            <nav data-testid="anchor-menu" className="nav" aria-label={menuTitle}>
+            <nav
+              data-testid="anchor-menu"
+              className="nav"
+              aria-label={menuTitle}
+            >
               {items?.map(item => (
                 // Use this package button
                 // @ts-ignore
--- a/packages/components-core/src/components/Pagination/PageItem/index.js
+++ b/packages/components-core/src/components/Pagination/PageItem/index.js
@@ -22,6 +22,7 @@ export const PageItem = ({
   ellipses,
   ariaLabel,
   children,
+  ariaDisabled,
 }) => {
   return (
     <li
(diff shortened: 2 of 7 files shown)


mlsamuelson (changes requested):
@davidornelas11 We need to model these changes as best as possible (it doesn't need to be interactive) in `unity-bootstrap-theme` too, so we keep that reference in sync.

davidornelas11:
> @davidornelas11 We need to model these changes as best as possible (it doesn't need to be interactive) in `unity-bootstrap-theme` too, so we keep that reference in sync.

So my thinking on this is that most of these changes need Javascript to take effect, so i can hard code some of the things like aria-label and aria-current, but we would need to make it known that if someone utilizes this in the bootstrap-theme form and not components-core, they need to make sure they have javascript taking care of this

mlsamuelson (changes requested):
Items we discussed that need to be addressed:

* Template story label needs to include "page" in the previous/next symbol's links' `aria-label` attributes - so "Previous page" and "Next page" instead of just "Previous" or "Next". (This of course should also be true for the example stories.)
* Example stories are missing `aria-label` on the previous/next symbols' links.
* Example stories' "previous" anchor tag has a `tabindex="-1"` on it, which makes it unfocusable.
* Let's add a README docs page in Storybook for the pager that says. "The pager stories here are static, markup-only references. Developers will need to provide interactivity that updates markup accordingly as users interact with the pager. Care must be taken to ensure accessibility. Please refer to the [W3C Pagination recommendations](https://design-system.w3.org/components/pagination.html) for best practices.

mlsamuelson (approved):

PR #1151, opened 2023-08-23
feat(unity-bootstrap-theme): new BG patterns

scott-williams-az:
#### New BG patterns
- Semiconductor
- Plus
- Arrows

### Links
- https://brandguide.asu.edu/brand-elements/design/patterns
- [JIRA ticket](https://asudev.jira.com/browse/UDS-1441)


packages/unity-bootstrap-theme/src/scss/variables/_dividers.scss
mlsamuelson, line 1: @scott-williams-az Recognizing that you were just following the established pattern here... but I'm thinking `_dividers.scss` doesn't make a lot of sense for these (and the original) variables. Are we agreed that `_backgrounds.scss` might be a better choice? Or am I missing something?

Otherwise, this all looks good.

mlsamuelson (changes requested):

mlsamuelson (approved):

PR #1153, opened 2023-08-24
fix(unity-bootstrap-theme): nested containers failsafe, include conta…

scott-williams-az:
nested containers failsafe, include container end styles only once

### Description
button issue on mobile was caused by nested container padding
https://chs.asu.edu/bachelors-degrees/majorinfo/ECEXERBS/undergrad/false/91407

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1230)

### Checklist

- [ ] Add/updated examples
- [ ] No new console errors
- [ ] Accessibility checks

### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Images

image

image

### DIFF
--- a/packages/unity-bootstrap-theme/src/scss/_end-style.scss
+++ b/packages/unity-bootstrap-theme/src/scss/_end-style.scss
@@ -1,0 +1,93 @@
+
+/* -----------------------------------------------------
+Container/ row / column padding adjustments for mobile.
+
+Bootstrap doesn't natively provide a way to alter the behavior
+its native grid elements based on a media query. This overrides
+that behavior at screens <= 575px. (The small breakpoint.)
+
+Should be included after @import scss/grid. Registered here for clarity.
+------------------------------------------------------ */
+@mixin end_styles() {
+
+  @include media-breakpoint-down(md) {
+    .container,
+    .container-fluid,
+    .container-xl,
+    .container-lg,
+    .container-md,
+    .container-sm {
+      padding-left: 2rem;
+      padding-right: 2rem;
+    }
+  }
+
+
+  .container .container,
+  .container .container-sm,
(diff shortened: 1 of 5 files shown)


packages/unity-bootstrap-theme/src/scss/_end-style.scss
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+
+  .container .container,
+  .container .container-sm,
+  .container .container-md,
+  .container .container-lg,
+  .container .container-xl,
+  .container .container-fluid,
```
scott-williams-az, line 31: https://dev-webspark-release-stable.ws.asu.edu/
Header has nested `.container .container-fluid` is this a valid case?

mlsamuelson: @scott-williams-az That should probably be corrected. Doesn't sound right.  Does this update trigger issues with that, or can it go forward and that issue be resolved after?

scott-williams-az: WS2 would need to be corrected first

mlsamuelson (changes requested):
Prior to merging this we need to have https://asudev.jira.com/browse/WS2-1775 addressed in WS2.

mlsamuelson:
@scott-williams-az I believe we decided this isn't advisable as the solution to this issue due to there being valid instances where nested containers are used in WS2. Right? Thoughts on an alternate approach?

scott-williams-az:
There are valid use cases for nested containers, however the nested container should have a parent with the row class. The solution should not be solved with css, we should ensure a nested container has a parent row going forward.

PR #1160, opened 2023-09-07
feat: added shared global variable func

davidornelas11:
### Description
Each component needs to have a global variable added with info about that component for easy tracking using tag manager. 

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1460?atlOrigin=eyJpIjoiNWRjOTU3MTg5MTdkNDgzODk3MGU5Nzc1ZGU0OTFmY2IiLCJwIjoiaiJ9)

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge


shared/functions/componentDatalayer.js
```diff
@@ -0,0 +1,32 @@
...
+ *   @param {Object} configuration - Configuration data for the component.
+ */
+function trackReactComponent({
+  componentName = "",
+  package = "",
+  component = "",
+  type = "",
+  configuration = {}
+}) {
+  if (!componentName || !package || !component || !configuration) {
```
scott-williams-az: Remove `configuration` check req another way

with the default `configuration = {}`, `!configuration` will always be `false` unless the user passes in a value like false (which wouldn't make sense)

shared/functions/componentDatalayer.js
```diff
@@ -0,0 +1,32 @@
...
+      console.error("trackReactComponent: Missing required arguments.");
+      return;
+  }
+
+  // Initialize or retrieve the global object to store component data.
+  window.udsComponent = window.udsComponent || {};
```
scott-williams-az: I'd like to start using this pattern more, what do you think of nesting further so we can use more things nested under one scope. `window.uds.dataLayer.component` and `window.uds.awesomeObjectOrFn`

shared/functions/componentDatalayer.js
```diff
@@ -0,0 +1,32 @@
...
+
+  // Store component data in the global object using the provided componentName.
+  window.udsComponent[componentName] = {
+      package: package,
+      component: component,
+      type: type,
```
scott-williams-az: I believe default empty string for `type` is OK, just checking.

shared/functions/componentDatalayer.js
davidornelas11: The only caveat of using this nested pattern is that the dataLayer must stay in the direct descendent window object for tag manager, i believe. But other than that I'm all for nesting the rest.

packages/app-rfi/src/components/AsuRfi/index.js
```diff
@@ -38,30 +40,48 @@ const AsuRfi = ({
...
+    if (typeof window !== "undefined") {
+      trackReactComponent({
+        packageName: "app-rfi",
+        component: "AsuRfi",
+        type: "NA",
+        configuration: {
+          campus,
+          college,
+          department,
+          studentType,
```
mlsamuelson: On the RFI component, tracking all configurations except for the `dataSource*` and `submissionUrl` props would be helpful. If it makes it easier, we could just rack the full set of props.

packages/component-footer/src/footer.js
```diff
@@ -18,6 +19,20 @@ const Base = () => {
  * @returns {JSX.Element}
  */
 const ASUFooter = ({ social, contact }) => {
+  useEffect(() => {
+    if (typeof window !== "undefined") {
+      trackReactComponent({
+        packageName: "component-footer",
+        component: "NA",
```
mlsamuelson: We should always have a component name.

packages/component-header/src/header.js
```diff
@@ -50,6 +51,21 @@ const ASUHeader = ({
     }
   };
 
+  useEffect(() => {
+    if (typeof window !== "undefined") {
+      trackReactComponent({
+        packageName: "component-header",
+        component: "NA",
```
mlsamuelson: We should have a component name.

mlsamuelson: Could we include a few more that are boolean true if present:
`navTree`
`parentOrg`
`buttons`
`mobileNavTree`

packages/component-news/src/components/CardGridNews/index.js
```diff
@@ -70,12 +71,30 @@ const GridTemplate = ({ cardButton }) => {
...
+const CardGridNews = ({ cardButton, ...props }) => {
+  useEffect(() => {
+    if (typeof window !== "undefined") {
+      trackReactComponent({
+        packageName: "component-news",
+        component: "CardGridNews",
+        type: "NA",
+        configuration: {
+          cardButton,
+          ...props,
```
mlsamuelson, line 83: Between `CardCarouselNews` and `CardGridNews` components the props are handled inconsistently - one uses spreading, the other doesn't. Let's be consistent.

packages/component-news/src/components/CardListlNews/index.js
```diff
@@ -67,12 +68,27 @@ const ListTemplate = ({ cardButton }) => {
...
-  </BaseFeed>
-);
+const CardListlNews = ({ cardButton, ...props }) => {
+  useEffect(() => {
+    if (typeof window !== "undefined") {
+      trackReactComponent({
+        packageName: "component-news",
+        component: "CardListlNews",
+        type: "NA",
+        configuration: {},
```
mlsamuelson: Missing `cardButton` and `props` to match `CardGridNews` and `CardCarouselNews` implementations.

shared/functions/componentDatalayer.js
mlsamuelson: Let's keep the `window.udsComponent` for this usage, since it needs to be a direct descendent for GTM to parse. But for all other UDS globals in the future, I agree with @scott-williams-az about nesting under `.uds` like so `window.uds.coolnessGoesHere`.

mlsamuelson (changes requested):
For `components-core` I’m thinking we don’t want to capture a global variable for those components  that would be used inside another component multiple times - like buttons, etc.  The spirit of what we want to track is "app-level" configs.  So global variable settings for button, card and testimonial components.... not the desired outcome.  For components core I think you could make a case for tracking the Pagination and TabbedPanels components... since there's likely to only be one on a page, but even then, maybe they're not necessary. ... `components-core` exists for other higher level components to use, so maybe we should not track anything from it.

shared/functions/componentDatalayer.js
mlsamuelson, line 1: Rename `shared/functions` to `shared/services`.  This will put the script alongside the googleAnalytics.js service.

mlsamuelson (dismissed):

shared/functions/componentDatalayer.js
mlsamuelson: @davidornelas11 Bumping this as a reminder. Just needs the path renamed.

mlsamuelson (approved):

PR #1175, opened 2023-10-04
fix(app-webdir-ui): focus cursor in search box on page load

davidornelas11:
### Description
Focus cursor in search bar on page load

### Links

- [Unity reference site](https://unity.web.asu.edu/@asu/app-webdir-ui/index.html?path=/story/organisms-search-page-templates--search-page-example)
- [JIRA ticket](https://asudev.jira.com/browse/SCHWEB-1089?atlOrigin=eyJpIjoiNDkxOGNiYmZkMTgzNGQ4NWFkNTc0NjM4Mzk2YTExYTIiLCJwIjoiaiJ9)

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### DIFF
--- a/packages/app-webdir-ui/.storybook/preview.js
+++ b/packages/app-webdir-ui/.storybook/preview.js
@@ -56,7 +56,7 @@ export const decorators = [
     const [args, updateArgs] = useArgs();
   return <MemoryRouter>
         <Wrapper args={args} updateArgs={updateArgs}>
-          <Story />
+          {Story()}
         </Wrapper>
     </MemoryRouter>
   },
--- a/packages/app-webdir-ui/src/SearchPage/index.js
+++ b/packages/app-webdir-ui/src/SearchPage/index.js
@@ -1,6 +1,6 @@
 import { Button, TabbedPanels, Tab } from "@asu/components-core";
 import PropTypes from "prop-types";
-import React, { useState, useEffect } from "react";
+import React, { useState, useEffect, ref } from "react";
 import { useSearchParams } from "react-router-dom";
 
 import { engineNames, engines } from "../helpers/search";
@@ -33,6 +33,7 @@ function SearchPage({
   const [searchParams, setSearchParams] = useSearchParams();
   const [filters] = useState({});
   const [site, setSite] = useState(null);
+  const inputRef = ref(null);
 
   const engineParams = {
     filters,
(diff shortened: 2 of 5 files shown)


mlsamuelson (changes requested):
Throws error in Storybook:
`(0 , react__WEBPACK_IMPORTED_MODULE_13__.ref) is not a function
TypeError: (0 , react__WEBPACK_IMPORTED_MODULE_13__.ref) is not a function`

Errors go away and the story renders if you import `useRef` instead of `ref` on line 3 and replace `ref` with `useRef` on line 36.

The focus doesn't correctly update, though. Because it looks like you're missing `ref={inputRef}` on the input tag, around line 131.

mlsamuelson (commented):
Updates look good, but see https://github.com/ASU/asu-unity-stack/pull/1177 which addresses the Storybook error "Rendered more hooks than during the previous render."

mlsamuelson:
The #1177 PR issue turns out to be unrelated to this PR. Just for the record.

mlsamuelson (dismissed):

scott-williams-az:
Do we need a way enable/disable focus onLoad? Have we identified the expected behavoir when there are 2 elements trying to focus on page load?
See Tabbed Panel https://github.com/ASU/asu-unity-stack/blob/dev/packages/components-core/src/components/TabbedPanels/components/TabHeader.js#L22

mlsamuelson:
@scott-williams-az Good point. I didn't realize that the tabbed panels were also capturing focus. And in the case of the Search Page component, it's using tabbed panels and the search focus...

scott-williams-az:
Now that I think about it, I may have changed the tabbed panels from `focus` to `scroll into view` (the code to `focus` on the tab might actually be unused, if it is we can remove it)

scott-williams-az:
Sorry for the back and forth, we are setting focus on the tabbed panels
https://github.com/ASU/asu-unity-stack/blob/dev/packages/components-core/src/components/TabbedPanels/index.js#L62

mlsamuelson:
I think that ought to work, and tests out locally.

mlsamuelson (approved):

PR #1179, opened 2023-10-09
fix(app-webdir-ui): pagination and results now account for filtered p…

davidornelas11:
…rofiles

### Description
Error where the pagination and displayed people were not taking into account the user-supplied profiles to filter out is now fixed

### Links

- [Unity reference site](https://unity.web.asu.edu/@asu/app-webdir-ui/index.html?path=/story/organisms-web-directory-templates--web-directory-example-departments-and-people)
- [JIRA ticket](https://asudev.jira.com/browse/SCHWEB-1031?atlOrigin=eyJpIjoiMjUxYTA0ZDYwZmE4NDA5YWE5YmM4MzMzZGRhN2RlYjgiLCJwIjoiaiJ9)
### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### DIFF
--- a/packages/app-webdir-ui/src/WebDirectoryComponent/index.stories.js
+++ b/packages/app-webdir-ui/src/WebDirectoryComponent/index.stories.js
@@ -98,7 +98,7 @@ export const webDirectoryExampleDepartmentsAndPeople = createStory(
       <div className="uds-content-align">
         <WebDirectory
           searchType="people_departments"
-          ids=",tgrandli:1344,mcrow:1343,jcunnin8:1358,ccherrer:1358,csmudde:1358"
+          ids="mcrow:1343,ccherrer:1358,csmudde:1358"
           API_URL="https://test-asu-isearch.ws.asu.edu"
           searchApiVersion="/api/v1/"
           display={display}
--- a/packages/app-webdir-ui/src/WebDirectoryComponent/index.js
+++ b/packages/app-webdir-ui/src/WebDirectoryComponent/index.js
@@ -33,8 +33,16 @@ function WebDirectory({
       tempFilters["deptIds"] = deptIds.split(",");
       return tempFilters;
     }
-    tempFilters["peopleInDepts"] = ids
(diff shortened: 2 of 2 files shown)


packages/app-webdir-ui/src/WebDirectoryComponent/index.stories.js
```diff
@@ -98,7 +98,7 @@ export const webDirectoryExampleDepartmentsAndPeople = createStory(
       <div className="uds-content-align">
         <WebDirectory
           searchType="people_departments"
-          ids=",tgrandli:1344,mcrow:1343,jcunnin8:1358,ccherrer:1358,csmudde:1358"
```
davidornelas11, line 101: Removed these as they're not valid asurite:department# pairs.

mlsamuelson (changes requested):
@davidornelas11 This fix doesn't resolve the issue completely. To replicate the issue, in the WebDirectoryComponent index.stories.js, replace lines 13-18 with the following arguments:
```
const display = {
  defaultSort: "last_name",
  doNotDisplayProfiles: "baaresta,iemxa",
  profilesPerPage: "3",
  usePager: "1",
};
```
Then visit the "Web Directory Example Departments" story. What happens, the first two profiles from the results (`baaresta,iemxa`, our doNotDisplayProfiles) are filtered out, and the first page will only have 1 result. Subsequent pages will have 3 users per page.

In the ticket, I'd suggested we solve this by sending the users to be filtered out through the endpoint so that Elastic can do the job of filtering the users out - and if we do that, the component doesn't have to implement any special logic around that - it can simply page through the results returned and we won't have empty spots that have been filtered out resulting in inconsistent numbers of profiles per page.

davidornelas11:
Closing this as I created another PR [here](https://github.com/ASU/asu-unity-stack/pull/1192)

PR #1182, opened 2023-10-13
docs(app-webdir-ui): added JSDOC type definitions

davidornelas11:
### Description
Added jsdoc type definitions

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1292?atlOrigin=eyJpIjoiZjkzMGIzZGZkYmE2NDk0OWEyZWM0NjRkOWMzY2RkZTQiLCJwIjoiaiJ9)

### Checklist

- [ ] Add/updated examples
- [ ] No new console errors
- [ ] Accessibility checks

### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge


mlsamuelson (commented):
Since the `docs/` folder ends up in the packages, I'd like to see if we can get them included.  I’m thinking if a call to the `docs` script happened in the `postbuild` script it would be output in /docs and included in the release.

That would leave committing the latest `docs/`  output to the repo as the responsibility of the developer, but since during the course of regular development the `postbuild` command would be triggered, we'd be fairly likely to have the latest and the developer would be prompted by that file showing as changed. That to me is preferable to complicating the build process for that piece. "Fairly likely" to be the latest is good enough, knowing that devs with the repo can do a `yarn docs` and get the latest as needed. My main concern is that package consumers have the latest version to match their package.

mlsamuelson (approved):

PR #1183, opened 2023-10-17
fix(app-webdir-ui): fix rendered more hooks than during previous render

mlsamuelson:
Need this fixed for presentation on Thursday. Was previously merged into #1175, but we don't want that effort blocking this getting into `dev`


davidornelas11 (approved):
Looks good!

PR #1189, opened 2023-10-20
Uds 1481

scott-williams-az:
### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1481)

### Checklist

- [x] Add/updated examples
- [ ] No new console errors
- [ ] Accessibility checks

### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge


mlsamuelson (commented):
When https://github.com/ASU/asu-unity-stack/commit/930de5829824601058bb1ef4e3b413d8356cba4e went in, I missed the implications that we were making `components-core` a peer dependency for the `component-header` instead of the necessary bits being compiled into the package.

`component-header` is designed to be standalone from other packages in that it doesn't require `unity-bootstrap-theme`. Current usage of `components-core` within the header and in this PR is for the utilities `getCurrentScriptPath` and `idGenerator`, so we're not mis-using it, but it has the possibility of opening the door to that. For this one case does it make sense as a peerDep?

Possible paths forward: 
1. Keep `components-core` in `component-header` as is (peerDependency) but update README.md for the package so the installation instructions spell this out more clearly and convey how to use the package - could be as simple as pointing out the need to provide the peerDeps. (Also review other peerDependencies for the package - are there any we want to highlight and/or move to the dependencies?).
2. Revert back, for this one package, to pulling in the utility scripts by path, so we have the power and convenience of code-reuse, but not the dependency on `components-core` for users to manage.
3. Other ideas?

mlsamuelson:
Shared functions folder is to be added in https://github.com/ASU/asu-unity-stack/pull/1160

mlsamuelson:
Informed by https://github.com/ASU/asu-unity-stack/pull/1160#pullrequestreview-1704711882

packages/component-header/package.json
```diff
@@ -37,7 +37,7 @@
     "build-storybook": "build-storybook -s ./dist"
   },
   "peerDependencies": {
-    "@asu/components-core": "^2.1.0",
+    "@asu/components-core": "^2.1.0 || ^3.0.0",
```
mlsamuelson: Remove `components-core` as peerDep. Move referenced code to `utilities/` and reference it by path so it gets compiled into package.

Also, let's remove peerDeps for `https://github.com/ASU/asu-unity-stack/pull/1160#pullrequestreview-1704711882` and `prop-types` and `styled-components`.

packages/component-footer/package.json
```diff
@@ -36,8 +36,10 @@
     "storybook": "start-storybook -s ./dist -p 9070",
     "build-storybook": "build-storybook -s ./dist"
   },
+  "peerDependencies": {
+    "@asu/components-core": "^2.1.0 || ^3.0.0"
```
mlsamuelson: Maybe include this by path instead, too.

mlsamuelson (changes requested):

scott-williams-az:
I ended up moving more scripts than we originally planned, and I am not set on the folder named "internal" so we can change that if needed.

mlsamuelson:
@scott-williams-az Let's change from `internal/` to `shared/`.

mlsamuelson (changes requested):
Seeing some errors. Maybe the `sanitizeDangerousMarkup` export needs to be uncommented on https://github.com/ASU/asu-unity-stack/pull/1189/files#diff-d54e7a9426b0944ee79ed9740d313a0c066c5980e42b6ed50d29b1a71619b2a7R7

Screenshot 2023-11-16 at 4 09 24 PM

mlsamuelson (approved):

PR #1192, opened 2023-10-24
fix(app-webdir-ui): add webdir profiles for exclusion on rest endpoint

davidornelas11:
### Description
The correct params and data were not being sent for the REST endpoint to properly filter out profiles in webdir

Added correct params.
This depends on having the correct login in place for the isearch repo. [Here](https://github.com/ASUWebPlatforms/asu-isearch/pull/542) is the pr for that

### Links

- [JIRA ticket](https://asudev.jira.com/browse/SCHWEB-1031?atlOrigin=eyJpIjoiZWY0MzAwMzJhYWFmNGJlNmE3ZDQ5MTkwZTNjZWU1NzgiLCJwIjoiaiJ9)

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### DIFF
--- a/packages/app-webdir-ui/src/helpers/search.js
+++ b/packages/app-webdir-ui/src/helpers/search.js
@@ -360,18 +360,14 @@ export const engines = {
     resultsPerSummaryPage: 6,
     supportedSortTypes: ["_score_desc", "last_name_desc", "last_name_asc"],
     method: "POST",
-    formatter: ({ results, page, itemsPerPage, cardSize, filters, appPathFolder }) => {
-      const startIndex = (page - 1) * itemsPerPage;
-      const endIndex = startIndex + itemsPerPage;
-      const resultsToDisplay = results.slice(startIndex, endIndex);
-
-      return webDirDeptsFormatter({
+    formatter: ({ results, cardSize, filters, appPathFolder }) =>
+      webDirDeptsFormatter({
         engineName: engineNames.WEB_DIRECTORY_PEOPLE_AND_DEPS,
-        results: resultsToDisplay,
+        results,
         cardSize,
         filters,
         appPathFolder,
-      })},
+      }),
     needsTerm: false,
   },
 };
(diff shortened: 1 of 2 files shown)


mlsamuelson:
Requires https://github.com/ASUWebPlatforms/asu-isearch/pull/535 to be approved, merged and deployed prior to us merging this.

packages/app-webdir-ui/src/helpers/search.js
```diff
@@ -458,15 +466,13 @@ export const performSearch = function ({
         "X-CSRF-Token": tokenResponse.data,
       };
       const data = {
-        "size":
-          filters.peopleInDepts < itemsPerPage
-            ? filters.peopleInDepts.length
-            : itemsPerPage,
-        "page": page || "",
+        "size": 500, // This is to retrieve all results and handle paging on the front end.
```
davidornelas11: We might be able to change this to any big number if we expect a people directory to have more than 500

davidornelas11: According to elastic's docs "By default, you cannot use from and size to page through more than 10,000 hits. This limit is a safeguard set by the [index.max_result_window](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#index-max-result-window) index setting. If you need to page through more than 10,000 hits, use the [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after) parameter instead." 
so we should be fine with very large directories when thinking about elastic's capabilities

mlsamuelson: @davidornelas11 Cool. Appreciate you doing the legwork to check into the docs. I found that the EdPlus dept has approx 700 people in it. ... Also, why do we need to do this? With the change made to the API, Elastic's App Search API is now handling removing people before processing the results, so we should be able to rely on it for paging without running into the side effect we are trying to solve here.

mlsamuelson (approved):
Tests out. Looks good. Thanks, @davidornelas11

PR #1207, opened 2023-11-07
SCHWEB-1128: fix(app-webdir-ui): fix zero value EID in path problem

tbutterf:
SCHWEB-1128

### Description
Because some profiles do not have an EID, and the database doesn't allow NULL values on that field, any profile without an EID is supposedly accessable at the path `/profile/0`. However, since there are multiple profiles with this problem, only one is accessible at that path. To mitigate this, I have made changes on the CMS side, but I also felt it would be good to check for zero-value EIDs on the component side, and if the EID has a value of 0 to fall back to creating the path for the profile by using the asurite ID instead (thus yielding a path like `/profile/djmahone` instead.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/SCHWEB-1128)

### Checklist

- [x] No new console errors
- [x] Accessibility checks

### DIFF
--- a/packages/app-webdir-ui/src/helpers/dataConverter.js
+++ b/packages/app-webdir-ui/src/helpers/dataConverter.js
@@ -277,7 +277,7 @@ export const staffConverter = ({
 
   // We use EID if it's available, otherwise we use the asurite_id.
   const profileURLBase = options.profileURLBase ?? "";
-  const asuriteEID = filledDatum.eid.raw
+  const asuriteEID = filledDatum.eid.raw && filledDatum.eid.raw !== '0'
     ? filledDatum.eid.raw.toString()
     : filledDatum.asurite_id.raw.toString();
   if (appPathFolder) {
@@ -334,7 +334,7 @@ export const studentsConverter = ({
   if (appPathFolder) {
     anonImg = `${appPathFolder}/img/anon.png`;
   }
-  const asuriteEID = filledDatum.eid.raw
+  const asuriteEID = filledDatum.eid.raw && filledDatum.eid.raw !== '0'
     ? filledDatum.eid.raw.toString()
     : filledDatum.asurite_id.raw.toString();
 


packages/app-webdir-ui/src/helpers/dataConverter.js
```diff
@@ -277,7 +277,7 @@ export const staffConverter = ({
 
   // We use EID if it's available, otherwise we use the asurite_id.
   const profileURLBase = options.profileURLBase ?? "";
-  const asuriteEID = filledDatum.eid.raw
+  const asuriteEID = filledDatum.eid.raw !== '0'
```
mlsamuelson: @tbutterf Couple issues here. 
1. It's only applying to the `staffConverter` and should also be implemented for the `studentsConverter`
2. You're regressing the logic on line 280 by replacing the original check with a different check. `filledDatum.eid.raw && filledDatum.eid.raw !== "0"` (untested) would be more appropriate.

mlsamuelson (changes requested):

packages/app-webdir-ui/src/helpers/dataConverter.js
tbutterf: Ah. Dang. Thanks. I'll update it.

davidornelas11: One more thing, can we always expect there to be an eid.raw? If this is certain, then we can keep it as it was before,  but if there could be cases where it is not available, optional chaining `filledDatum?.eid?.raw` might be a good addition. This was already present before this PR, but might be a nice addition since it's already getting changed. With the switch over to a different people management software, would the data be changed at all? @mlsamuelson

tbutterf: I have updated it according to your notes @mlsamuelson. I didn't add the question marks that @davidornelas11 suggested, but I can, if needed.

davidornelas11: I don't think it's important as long as the data will not change, and if it does we can always update the code

mlsamuelson: The data could change at some point, and I always appreciate code that's guarded against a rainy day scenario, but two mitigating circumstances on this:
1. We've been okay thus far, and checking the returns from our API, looks like if the field is returned, raw is always included, and Workday is a way off and will involve testing before we roll out changes.
2. We're about to put the data converters under the microscope with the code review, so there may be changes coming to this area, and/or we can identify this change as an outcome of that effort. I'm guessing there are more opportunities to address the raw issue throughout the data converters, and if we make that DRY, we can tackle all at once, perhaps.

Thus I'm not going to have this hung up on that detail. :)

mlsamuelson (approved):

PR #1218, opened 2023-11-29
Uds 1531 - add props and code to Image component to match Bootstrap version

davidornelas11:
### Description
Add props to match the multiple versions of Image component used in unity-bootstrap-theme.
https://github.com/ASUWebPlatforms/webspark-ci/pull/214 is dependent on this pr

### Links

- [Unity reference site](https://unity.web.asu.edu/@asu/components-core/index.html?path=/story/uds-image--default)
- [JIRA ticket](https://asudev.jira.com/browse/UDS-1531?atlOrigin=eyJpIjoiNzg5YTUxYTVjNTE1NDgxZmJkNzg0MjNmNzkzNmZhYTMiLCJwIjoiaiJ9)

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### DIFF
--- a/packages/components-core/src/components/Image/index.js
+++ b/packages/components-core/src/components/Image/index.js
@@ -48,17 +48,19 @@ export const Image = ({
     "uds-img-drop-shadow": dropShadow,
   });
 
-  const renderImage = () =>
+  const renderImage = (classes) => {
+    let combinedClasses = classes ? `${imageProps.className} ${classes}` : imageProps.className;
+  return (
     cardLink ? (
       <a href={cardLink}>
         {/* eslint-disable-next-line jsx-a11y/alt-text, react/jsx-props-no-spreading */}
-        <img {...imageProps} />
+        <img {...imageProps} className={combinedClasses} />
         <span className="visually-hidden">{title}</span>
       </a>
     ) : (
       // eslint-disable-next-line jsx-a11y/alt-text, react/jsx-props-no-spreading
-      <img {...imageProps} />
-    );
+      <img {...imageProps} className={combinedClasses} />
+    ))};
 
   const renderFigure = () => (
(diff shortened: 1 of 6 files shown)


mlsamuelson (changes requested):
Code is looking good, but I've found some regressions in testing, and some pieces we need to seek clarification on:

1. The img's all have a border of 1px on them now. It's easy to see this one if you look at the default card stories or the event card. This appears to be because we're now using the `uds-img` on the wrapper div. Let's be sure we understand what that's doing before changing the behavior there so we don't regress elsewhere.
2.  Story cards and News Story cards and horizontal story cards have regressed. (No need to describe, it's pretty visible. )
3. Captions... We have some things to work through here. These are essentially the same component, but the XD references disagree on Caption titles and text color.  Differences in captions between Carousel image cards and Images: 
  a. Carousel image cards caption text allows titles - but there are no examples of titles in captions for images. > Need to determine if this acceptable.
  b. Carousel image cards captions text and titles are Gray7. Images with text in the XD are Gray5. > the Gray5 applied to cards with caption titles looks bad. ... can we standardize on Gray7 for all cases or do we need to code for variations based on use?
  c. On captions in images, the margins don't match margins used in Carousel images.
4. Not sure if this is because of this update, but in testing, I noticed the Ranking card transition effect is more choppy now.

Note: Some of these changes also impact Ranking Cards due to style reuse, so please check them for regressions with any updates.

mlsamuelson:
From post-scrum: @imorale2 shared that because the XD isn't fully up to date, the Image Carousel image cards should be considered our canonical example. So that means:
* font color should be Gray7
* titles are allowed

mlsamuelson (approved):

PR #1222, opened 2023-12-04
fix(components-core): allow sanitized html in image citation

mlsamuelson:
WS2-1711 requires the Image component's citation to allow basic, sanitized HTML. Doing that here.


davidornelas11 (approved):
Tested out and looks good!

PR #1228, opened 2024-01-11
feat(app-rfi): use DPL  DataSource

scott-williams-az:
### Links

- [Unity reference site](https://unity.web.asu.edu/@asu/app-rfi/index.html)
- [JIRA ticket](https://asudev.jira.com/browse/ERFI-136)

### Checklist

- [ ] Add/updated examples
- [ ] No new console errors
- [ ] Accessibility checks

### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge


packages/app-rfi/src/components/stepper/RfiStepper.js
```diff
@@ -27,28 +30,41 @@ const RfiStepper = props => {
...
-          .then(response => response.json())
-          .catch(error => new Error(error));
-        // Structure of response: response.programs[0].EmailAddr
-        if (res.programs) {
-          const {
-            programs: [{ EmailAddr }],
-          } = res;
-          setEmail(EmailAddr);
-        }
+      // Do we need this isCertMinor prop, or can we read the data to decide
```
mlsamuelson, line 53: A good question. With the new service it may be easier to determine reliably when something is a cert or minor...

packages/app-rfi/src/components/stepper/RfiStepper.js
```diff
@@ -27,28 +30,41 @@ const RfiStepper = props => {
...
+      if (isCertMinor && programOfInterest) {
+        fetchDegreesData({
+          dataSourceDegreeSearch,
+          programOfInterest,
+        }).then(([response, data]) => {
+          if (response === "Error") {
+            return;
+          }
+          const { emailAddr, planType } = data[0];
+          // debugger;
```
mlsamuelson: Remove?

packages/app-rfi/src/components/stepper/RfiStepper.js
```diff
@@ -70,7 +86,7 @@ const RfiStepper = props => {
   // See Formik validation flavors: https://formik.org/docs/guides/validation
   const validate = values => {
     const errors = {};
-    const { programOfInterest, programOfInterestOptional } = props;
+    // const { programOfInterest, programOfInterestOptional } = props;
```
mlsamuelson, line 90: Remove.

packages/app-rfi/src/components/stepper/RfiStepper.js
davidornelas11: I have a check in the degree pages data that checks if the degreeType is UGCM, since from I understand, that covers if it's a certificate or minor.

packages/app-rfi/src/components/stepper/RfiStepper.js
```diff
@@ -27,28 +30,41 @@ const RfiStepper = props => {
...
-          } = res;
-          setEmail(EmailAddr);
-        }
+      // Do we need this isCertMinor prop, or can we read the data to decide
+      if (isCertMinor && programOfInterest) {
+        fetchDegreesData({
+          dataSourceDegreeSearch,
+          programOfInterest,
+        }).then(([response, data]) => {
+          if (response === "Error") {
```
davidornelas11, line 59: Since there are 2 types of errors in your fetchDegree service, an error if there is no data and an error with the dpl api itself, should we differentiate between errors here? Just brainstorming, since we might not need to.

davidornelas11: Also, should we add a signal to cancel any fetch requests to avoid memory leaks?

davidornelas11 (commented):
Not too much to add besides some brainstorming and questions, this code looks great! 👍

packages/app-rfi/src/components/stepper/RfiStepper.js
scott-williams-az: I think you bring a good point that we might get different errors and we would have to dig to find what the error was, but I'd like to keep the return value as the tuple pattern so typescript knows this function will always return an array of ["Error", errorObject] or ["Success", programData]

That way in a function if we have ruled out the response is an not error, Typescript Intelesense knows the response object has the expected AcadPlan shape. 

How do you feel about changing the code to actually log the error so we know what kind of error happened.

scott-williams-az: on the topic of fetching data. I did not know when I started this ticket that this file existed `shared/hooks/use-fetch.js` however this function does not have a way to cancel the request either. 

I like the idea of improving how we fetch data

packages/app-rfi/src/components/stepper/RfiStepper.js
mlsamuelson: @davidornelas11 UGCM works for undergrad certs and minors (the UG), but have you confirmed that with the grad certs? This is jogging my memory. I believe it was the lack of reliable flag for grad certs that led to the current business logic. cc @scott-williams-az

packages/app-rfi/src/components/stepper/RfiStepper.js
mlsamuelson: There's a lot talked about here, so let's make it actionable:

I like the idea of the code logging the error (assuming we're talking console.log() and not the addition of more complicated infrastructure :D ). That seems reasonable to include in this effort/PR.

@scott-williams-az Let's please decouple the larger task of improving how we fetch data. Please create a ticket for that. Perhaps as a first step, let's you, @davidornelas11 and I meet to discuss requirements for that ticket. Send an invite for that, if you would.

packages/app-rfi/src/core/utils/fetchPrograms.js
```diff
@@ -0,0 +1,112 @@
+// Fetch degree data from Degree Search REST API.
+// Returns the full degree dataset for the given program (undergrad | graduate).
+// We do filtering in useEffect()s to manage filtering logic for our various
+
+import { KEY } from "./constants";
+import { normalizeDegreeData } from "./datasource-helper";
+
+// field options.
```
mlsamuelson: Looking like this comment line got detached from line 3.

mlsamuelson (commented):
First run through the code looks good. I've added a few comments. Going to use the updates and refactored bits to guide testing and validation of the component now.

mlsamuelson:
@scott-williams-az In side-by-side testing of this update against the latest static site build, I'm noticing discrepancies in Areas of Interest (and by extension Programs of Interest) when using the `college` prop. I'm suspecting that has to do with the new partial ownership concept related to degrees and the normalizing here: https://github.com/ASU/asu-unity-stack/pull/1228/files#diff-5c91b3486d623bfbf07bec87e666de652938c9f1e9586278468fc15a581b4f4dR37 because between the old and new approach, I'm seeing more coming back - which I'd expect with partial ownership now. When using the college code `CLW` however, I get the following Areas of Interest
* Law, Justice and Public Service
* Social and Behavioral Sciences

In the update, we get four Areas of Interest but only the first from the list above is among them. The Programs of Interest options are also vastly different. I'd like to account for this difference and determine if we need to reach out to the Degree Search team for confirmation. It may be that with our update we need to document that use of the `college` prop will be affected and site owners need to adapt - i.e. we add details to the release notes when this ships with WS2.

Still testing, but another item I've noted is a regression on the Certs/Minors display. The `Request information on this program by sending an email to `... message isn't displaying in the Cert story.

packages/app-rfi/src/components/steps/AboutMe.js
```diff
@@ -80,26 +91,26 @@ const AboutMe = () => {
...
-                }`,
-                text: termValue[1].split(":")[0],
+            const termData = data[0].applicationDeadlines
+              ?.sort((a, b) => (a.strm > b.strm ? 1 : -1))
+              .map(({ strm, strmDescription }) => ({
+                key: strm,
+                value: strm,
+                text: strmDescription,
               }));
             // Dedupe based on object key property as dupe terms can occur due
```
scott-williams-az: @mlsamuelson I think this is no longer needed, what do you think

packages/app-rfi/src/core/utils/dataSnapshot.json
scott-williams-az, line 1: This was helpful at the beginning, but I am questioning if we really need this

scott-williams-az:
@mlsamuelson @davidornelas11 
I am curious if adding a snapshot of options would be helpful or if it is "information overload" here is a screenshot.

- I don't think it is appropriate to spam the API and populate Storybook fields dynamically.
- Possible idea could be to populate latest options with a script we would run manually
- if the SB control is a dropdown, and the desired option is not in the list, user would need to alter the URL to insert custom value (is that an acceptable compromise?)
image

mlsamuelson:
Yeah, that is a lot of information to present. Sounds like there's two things on the table:
1. to present the options in the description or not
2. populating the controls with the options

For 1, what about simply displaying the DPL API URL where the user can refer to for options. And then for 2, keeping it simple and have it just be a string field. This way we don't deal with snapshot drift and devs having to manually alter urls to test... it's a "keep it simple" approach.

packages/app-rfi/src/components/steps/AboutMe.js
mlsamuelson: If the new service is not delivering duped data, then we don't need it and it can be removed. :)

packages/app-rfi/src/core/utils/dataSnapshot.json
mlsamuelson: Nope. We don't need it in the repo.

mlsamuelson (approved):

PR #1230, opened 2024-01-16
fix(component-header):highlight nav parent if subitem active

scott-williams-az:
### Description

Has anyone noticed on their site that if they have multiple columns in their menu items that anything not in the first column will not highlight the menu navigation when they are on that page?

### Links

- [Unity reference site](https://unity.web.asu.edu/@asu/component-header/index.html?path=/story/uds-asu-header--with-menu-columns)
- [JIRA ticket](https://asudev.jira.com/browse/UDS-1518)

### Checklist

- [ ] Add/updated examples
- [ ] No new console errors
- [ ] Accessibility checks

### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Images

image

### DIFF
--- a/packages/component-header/src/components/HeaderMain/NavbarContainer/NavItem/index.js
+++ b/packages/component-header/src/components/HeaderMain/NavbarContainer/NavItem/index.js
@@ -123,6 +123,8 @@ const NavItem = ({ link, setItemOpened, itemOpened }) => {
     }
   };
 
+  const activeChild = link.items?.flat().find(({ selected }) => selected);
+
   return (
     <NavItemWrapper
       // @ts-ignore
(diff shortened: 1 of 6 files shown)


scott-williams-az:
This could use some discussion, as I don't necessarily think the additional prop for currentPage is needed 

and the change in the first commit actually doesn't do anything (at least for WS2) since we are never setting a sub menu item with selected=true

mlsamuelson (approved):
The update looks good. 😄  At first I was concerned about cases where there was a domain in the menu item’s link path, but then I realized that would be for an off-site page and it wouldn’t be a page to trigger and “active” state.  Now, there may be cases where someone has used a full URL in menus, but if that’s the case, the solution will be: “don’t”. And if there’s some esoteric reason where that solution doesn’t work, we’ll address it when it arises. This moves us in the right direction and we'll address nuances if/when they arise.

PR #1233, opened 2024-01-25
fix(component-header): selected prop only on top level navTree items

scott-williams-az:
### Description
Originally we only supported selected prop on the top level navigation items. Since sub items still do not need this property I removed code that I introduced to check for that property on sub items.

I did discover that WS2 does incorrectly send items as an empty string, but I have added comments and code to handle this


packages/component-header/src/components/HeaderMain/NavbarContainer/NavItem/index.js
```diff
@@ -123,8 +124,6 @@ const NavItem = ({ link, setItemOpened, itemOpened }) => {
     }
   };
 
-  const activeChild = link.items?.flat().find(({ selected }) => selected);
```
scott-williams-az, line 126: If we want to support this pattern, we could still, and should be able to reuse a function from active-page
ex:
`const activeChild = hasActivePage(link.items)`

mlsamuelson (approved):

PR #1238, opened 2024-02-09
fix(unity-bootstrap-theme): adjust endorsed-logo css

scott-williams-az:
combine media queries and use css3 variable to implement

### Links

- [component-footer](https://unity.web.asu.edu/@asu/component-footer/index.html?path=/story/uds-asu-footer--unit-logo)
- [unity-bootstrap-theme](https://unity.web.asu.edu/@asu/unity-bootstrap-theme/index.html?path=/story/organisms-global-footer-examples--zero-columns-example&args=header:false)
- [JIRA ticket](https://asudev.jira.com/browse/UDS-1455)

### Checklist

~- [ ] Add/updated examples~
~- [ ] Add/updated READMEs/docs~
~- [ ] No new console errors~
~- [ ] Accessibility checks~

### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### DIFF
--- a/packages/unity-bootstrap-theme/src/scss/extends/_globalfooter.scss
+++ b/packages/unity-bootstrap-theme/src/scss/extends/_globalfooter.scss
@@ -99,16 +99,16 @@ footer {
 // #endorsed-footer, mobile first design
 
 #endorsed-logo {
+  --endorsed-logo-display: block;
+  --endorsed-logo-padding-y: 12px;
   --endorsed-logo-img-max-height: 100%;
   --endorsed-logo-img-width: 100%;
-  --endorsed-logo-padding-y: 12px;
-  --endorsed-logo-display: block;
   @include media-breakpoint-up(md) {
     & {
+      --endorsed-logo-display: flex;
       --endorsed-logo-padding-y: 20px;
       --endorsed-logo-img-max-height: 6rem;
       --endorsed-logo-img-width: auto;
-      --endorsed-logo-display: flex;
     }
   }
   padding-top: var(--endorsed-logo-padding-y);
@@ -152,8 +152,6 @@ footer {
 
 // #endorsed-footer, up to desktop
 @include media-breakpoint-up(md) {
-
-
   #social-media .nav {
     padding: $uds-size-spacing-5 0;
     justify-content: flex-end;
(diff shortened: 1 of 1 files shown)


scott-williams-az:
@davidornelas11 @mlsamuelson  What do you think of this css3 variable pattern. I think it can help in a lot of ways, but if we do not have a clear pattern it may end up confusing us. So I am open to ideas. @media query changes the var values and does not add any new rules. 

WS2 or other platforms can then use the variable if needed, and will not have to use `!important`

So in this PR code, css variables are at the root style and I am trying to be specific about how the variables are named
```
#endorsed-logo {
  --endorsed-logo-display: block; // is meant for endorsed-logo attribute display
  --endorsed-logo-img-width: // is meant for endorsed-logo  img attribute width
```

Can you think of a other things we should lookout for or think about. We don't want css variable to start conflicting.

mlsamuelson:
@scott-williams-az This pattern makes sense to me. I wonder about how best to surface this (or at least the functionality it enables) to consumers. Perhaps simply socializing it when possible. AND of course it would be a candidate for being a documented pattern under the https://asudev.jira.com/browse/UDS-1586 effort.

The naming convention is clear, though what do you think about avoiding using `-y` and instead being more explicit and granular with `-top` and `-bottom` in an attempt to align more closely with the naming convention of `selector-property`? Might make it more of a no-brainer what to name the variables.

I would like to get @bebaps 's opinion on it as well, as he's a good representative power-user of the theme.

bebaps:
@mlsamuelson @scott-williams-az I do agree with this approach and the use of CSS Variables in general, where it makes sense. I don't have all the context here, but I am guessing that the core discussion is around the naming pattern and not the use of the variables themselves. On that front, I do like this pattern, and I also agree that we should not use the shorthand `-y` but use `top|bottom|left|right`. If this PR will essentially set the precedent for the use of the variables and its naming, then I agree with this.

bebaps (dismissed):
@mlsamuelson @scott-williams-az I do agree with this approach and the use of CSS Variables in general, where it makes sense. I don't have all the context here, but I am guessing that the core discussion is around the naming pattern and not the use of the variables themselves. On that front, I do like this pattern, and I also agree that we should not use the shorthand `-y` but use `top|bottom|left|right`. If this PR will essentially set the precedent for the use of the variables and its naming, then I agree with this.

mlsamuelson (changes requested):
Thanks, @bebaps . Your review was in the spirit I was hoping for.

@scott-williams-az Can you change the use of `-y` as noted in previous comments. After that, this looks good to go.

mlsamuelson (approved):

PR #1265, opened 2024-04-10
UDS-1600 - Update Ranking cards for accessibility

davidornelas11:
### Description
Updated the functionality of the ranking cards to match the [original user experience document](https://docs.google.com/presentation/d/1v0tjxBInfFOrl56R_w6gTwI80MhOQpdiLRAogW0qF0Q/edit#slide=id.g1212f04cf8e_0_19) as well as update accessibility

### Links

- [Unity reference site](https://unity.web.asu.edu/@asu/components-core/index.html?path=/story/uds-ranking-card--large)
- [JIRA ticket](https://asudev.jira.com/browse/UDS-1600?atlOrigin=eyJpIjoiNjMwYTBjNjA3ODZlNGRkZGFhYWJlMzRhYWM1OTE5ZjAiLCJwIjoiaiJ9)
- [Unity Design Kit](https://xd.adobe.com/view/7764cadf-bdfd-4e37-b1ee-b2f8bb2d1694-2237/screen/ec1af606-9e31-4593-bedb-9168e54766cc/specs/)
- [Original ui/ux docs](https://docs.google.com/presentation/d/1v0tjxBInfFOrl56R_w6gTwI80MhOQpdiLRAogW0qF0Q/edit#slide=id.g1212f04cf8e_0_19)

### FOR APPROVERS

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge


packages/components-core/src/components/RankingCard/index.js
```diff
@@ -58,44 +58,71 @@ CitationWrapper.propTypes = {
...
+                id="dispatch"
+                className="btn-expand"
+                aria-label="Expand ranking"
+                type="button"
+                aria-expanded={open}
+                aria-controls={uniqueIdRef.current}
+              >
+                <h4>{heading}</h4>
+              </button>
+              <i className="fas fa-chevron-up" />
```
mlsamuelson: The `fa-chevron-up` `i` tag should go inside the button, like with the small size... as it is here, the chevron isn't clickable on the larger version and that's problematic.

mlsamuelson (changes requested):

packages/components-core/src/components/RankingCard/index.js
mlsamuelson: Screenshot 2024-06-10 at 3 30 34 PM

packages/components-core/src/components/RankingCard/index.js
```diff
@@ -58,44 +58,71 @@ CitationWrapper.propTypes = {
 
 const InfoLayerWrapper = ({ imageSize, body, heading, readMoreLink }) => {
   const [open, setOpen] = useState(false);
+  const uniqueIdRef = useRef(
+    `info-layer-${Math.floor(Math.random() * 100000)}`
```
scott-williams-az: replace math.random `useRef` and `Math.random` with React `useId`
I don't think we need `useRef`, aria control works with just ID.
Example in links use aria control in a ver similar way
checkout https://react.dev/reference/react/useId#usage

davidornelas11: Scott's recommendation is definitely the way to go, but from my understanding useId is only available in [react 18](https://react.dev/blog/2022/03/29/react-v18#new-hooks) right?

scott-williams-az: OK, leave a comment to replace to useId when we get to v18. but you still may not need `useRef`

mlsamuelson: Yeah, that matches my understanding of when that feature is available. Also agree with the comment for future improvement. Good stuff.

mlsamuelson (changes requested):
The clickability of the title in the large ranking cards differs in `components-core` vs `unity-bootstrap-theme`.

mlsamuelson (approved):
Looks like the issues with this PR is that the markup updates were only made to the `components-core` versions of the stories. I’m going to go ahead and merge this, since WS2 is blocked and it only uses the `components-core` version of that card. And you can please follow up under the UDS-1720 PR ticket with a separate PR to update the theme story markup for Ranking Cards.

PR #1267, opened 2024-04-15
fix(app-degree-pages): added professional licensure section in degree…

davidornelas11:
… detail page

### Links

- [Unity reference site](https://unity.web.asu.edu/@asu/app-degree-pages/index.html?path=/story/program-detail-page--default)
- [JIRA ticket](https://asudev.jira.com/browse/UDS-1635?atlOrigin=eyJpIjoiZGU0YTI0YzI4NmUxNDU4Njg2NzRlOWE5NjFiMWUwYjQiLCJwIjoiaiJ9)

### FOR APPROVERS

### Checklist

- [] Commits do not contain multiple scopes
- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### DIFF
--- a/packages/app-degree-pages/src/core/services/degree-data-prop-resolver-service.js
+++ b/packages/app-degree-pages/src/core/services/degree-data-prop-resolver-service.js
@@ -138,6 +138,7 @@ function degreeDataPropResolverService(row = {}) {
     getEmailAddress: () => row["emailAddr"],
     /** @return {string} */
     getPhone: () => row["phoneNumber"]?.replace("/", "-"),
+    getProfessionalLicensureText: () => row["professionalLicensureAdditionalText"] || row["professionalLicensureStandardText"],
     /** @return {string} */
     getDepartmentName: () => getMajorityOwner(row)?.departmentDescription,
     /** @return {string} */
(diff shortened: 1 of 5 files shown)


packages/app-degree-pages/src/core/services/degree-data-prop-resolver-service.js
```diff
@@ -138,6 +138,11 @@ function degreeDataPropResolverService(row = {}) {
     getEmailAddress: () => row["emailAddr"],
     /** @return {string} */
     getPhone: () => row["phoneNumber"]?.replace("/", "-"),
+    getProfessionalLicensureText: () => {
+      const standardText = row["professionalLicensureStandardText"] || "";
+      const additionalText = row["professionalLicensureAdditionalText"] || "";
+      return `${standardText}${additionalText}`;
+    },
```
davidornelas11: Additional text is not present on all degrees that have the professional standard text. This combines both fields and has a fallback if there is no additional text. According to the ticket, the standard field may have "malformed html". If we want to validate the html, we could use jsdom to validate it before it gets added. There is also a native web api, DOMParser, available in all browsers so we can use this if we didn't want to add another library, but it is quite forgiving and will parse some invalid html

mlsamuelson: For historical record, we discussed and decided to go with the DOMParser approach for now. We prefer to have data issues solved at the data source, rather than have our  consumers implement heavy measures to address other's issues.

mlsamuelson (approved):

PR #1294, opened 2024-06-10
fix(component-header): fixed form submit null issue

davidornelas11:
setTimeout added a closure issue. Fixed by defining the event and added function into react component

### Description
setTimeout added a closure issue. Fixed by defining the event and added function into react component

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1719?atlOrigin=eyJpIjoiYmUxZTk2YTA0Y2VjNDM5NGExNzg5ZWZiZmI1MTFiYTMiLCJwIjoiaiJ9)

### FOR APPROVERS

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### DIFF
--- a/packages/component-header/src/components/UniversalNavbar/Search/index.js
+++ b/packages/component-header/src/components/UniversalNavbar/Search/index.js
@@ -1,4 +1,3 @@
-// @ts-check
 import { faSearch, faTimes } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import React, { useState, useRef, useEffect } from "react";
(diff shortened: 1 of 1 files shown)


packages/component-header/src/components/UniversalNavbar/Search/index.js
```diff
@@ -48,8 +26,32 @@ const Search = () => {
...
+      text: e.target.elements.q.value,
+    });
+    setTimeout(() => {
+      form.submit();
+    }, 100);
+  };
+
   const handleChangeVisibility = () => {
-    setOpen(prevState => {
+    setOpen((prevState) => {
```
scott-williams-az, line 54: Approved with or without this comment. Are our prettier setting somehow different? Single props do not get wrapped in () or should we update prettier settings?

scott-williams-az (approved):

packages/component-header/src/components/UniversalNavbar/Search/index.js
mlsamuelson: Good find. Let's not consider this blocking this issue since it itself is a blocker. Can be resolved elsewhere.

PR #1299, opened 2024-06-19
fix(component-header): added default GTM info to link clicks

davidornelas11:
### Description
added default GTM info to link clicks

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1726?atlOrigin=eyJpIjoiODliOThiMDdkZTI4NGU4MzhmNjA3ZWEzYTYwNGJlMGMiLCJwIjoiaiJ9)

### FOR APPROVERS

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

### DIFF
--- a/packages/component-header/src/components/HeaderMain/NavbarContainer/NavItem/index.js
+++ b/packages/component-header/src/components/HeaderMain/NavbarContainer/NavItem/index.js
@@ -24,6 +24,7 @@ export const LINK_DEFAULT_PROPS = {
   action: "click",
   name: "onclick",
   type: "internal link",
+  region: "navbar",
   section: "main navbar",
   text: "",
 };
@@ -103,6 +104,7 @@ const NavItem = ({ link, setItemOpened, itemOpened }) => {
       // @ts-ignore
       isDropdown
         ? {
+            ...LINK_DEFAULT_PROPS,
             ...DROPDOWNS_GA_EVENTS,
             action,
             text,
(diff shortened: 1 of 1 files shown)


mlsamuelson (changes requested):
I think we're on track with this. I noticed in the CFO site, there's a `region` value that's not being set in the fix yet, and the fix includes a `component` value that's not set (and also not in the CFO site). That `component` value - was that one we're setting as part of the global variable update for tracking components in the GTM? I didn't think we'd be touching the dataLayer with that... And maybe it's not that.

The other thing I noticed is that there aren't a full set of values being sent in for the `collapse` event clicks on the menu for the drop downs like there are on the pre-regressed CFO site.

So I think we still need to address those two items:
* `region` and maybe `component`
* `collapse` event clicks need full event details too.

davidornelas11:
> I think we're on track with this. I noticed in the CFO site, there's a `region` value that's not being set in the fix yet, and the fix includes a `component` value that's not set (and also not in the CFO site). That `component` value - was that one we're setting as part of the global variable update for tracking components in the GTM? I didn't think we'd be touching the dataLayer with that... And maybe it's not that.
> 
> The other thing I noticed is that there aren't a full set of values being sent in for the `collapse` event clicks on the menu for the drop downs like there are on the pre-regressed CFO site.
> 
> So I think we still need to address those two items:
> 
> * `region` and maybe `component`
> * `collapse` event clicks need full event details too.

From what I understand, the component value was set a few years ago and from what I found, it was never really used but I think adding that in the shared function might be better.  Never mind, it is set in the shared function so that’s still there. The region was hardcoded as well so that will need to be hard coded again

mlsamuelson (approved):

PR #1302, opened 2024-06-24
Update title logic webdir

davidornelas11:
### Description
Title logic matches new logic required by provost

### Links

- [JIRA ticket](https://asudev.jira.com/browse/SCHWEB-1238?atlOrigin=eyJpIjoiZTk0Njc3MzEzZDY5NDJmYWIyODlhZTRhOTlkMWM5MmEiLCJwIjoiaiJ9)
- [Unity Design Kit](https://xd.adobe.com/view/f0f3543e-7b95-4de0-84c6-4b473b44463f-a5a0/screen/192765ba-2e8f-4823-9903-3476abdfe267/)

### FOR APPROVERS

### Checklist

- [x] Add/updated examples
- [x] No new console errors
- [x] Accessibility checks

### Browsers

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge


packages/app-webdir-ui/src/helpers/dataConverter.js
```diff
@@ -105,106 +105,97 @@ const fillInBlanks = datum => {
...
-  // Note on title logic
-  // - Engine WEB_DIRECTORY_PEOPLE_AND_DEPS supplies results with title logic
-  //   already handled by the service.
-  // - Engine WEB_DIRECTORY_DEPARTMENTS needs a little help with titles,
-  //   handled below using titleMatch.depts.
+function getTitleAndDeptFromIndex(profile, index) {
+  let title = profile.titles?.raw[index];
+  let dept = profile.departments?.raw[index];
+  if (profile.title_source.raw[index] !== "titles" && profile.working_title?.raw[0]) {
+    title = profile.working_title.raw[0];
```
mlsamuelson: This is doing a fallback to `working_title` when the `title_source` doesn't identify a custom title. ... This would only be okay as long as it's only used when there's no dept provided but because we have the index from the department here... 

I only see it being used for the Courtesy Affiliate logic flow... if this is unique to that case, It would be better to change the name of the function or to embed the logic in the code, otherwise this runs the risk of causing confusion at a later date.

BUT, after setting up a test in the Test environment, and setting the `WebDirectoryExampleDepartments` story so it uses `deptIds="1342"` (ASU department) it display's Audrey's record - and this is not her primary affiliation, but it's displaying her HR title (working title) while it should display no title. So more work is needed to ensure we match the logic outlined in https://docs.google.com/document/d/1ypPuf-zmEPpKLpmry3paQTsQ5uaxX-_l1IjCHANwykU/edit .

mlsamuelson (changes requested):

packages/app-webdir-ui/src/helpers/dataConverter.js
mlsamuelson: That's convoluted. In essence, when we're displaying a profile and the department is supplied, we only should fallback to the working title if the deptId matches the person's primary_deptid. ... There may be some wiggly-ness there, so maybe we can touch base with @tbutterf and align on approach.

mlsamuelson (changes requested):
[removed] comment posted on the wrong pr

mlsamuelson:
As decided in our work session:
* private method names to be improved.
* Remove old logic that was left in place to guard against null values coming back. When the service is updated, all titles can be sourced from the fields from the service, regardless of if the profile is Courtesy Affiliate or not - the API will own all the logic.

mlsamuelson (changes requested):
@davidornelas11 Found an issue and we need to determine if it's a with the service or the component - I suspect the latter. What's happening: all Search Page story profile results - whether on the main tab, or Faculty and Staff - lack the title and department. ... I haven't dug into the code yet, but what I suspect is happening is that the code flow for search should also be consulting the title and department fields from the service and displaying the results, but just without a department passed in the argument (or "unaffiliated" being the value, I think is what you'd pass) - the expectation in that case would be you'd get the expected primary title and affiliation back.

mlsamuelson:
Ahh, I think it's behaving pretty much correctly, except the endpoint isn't returning anything when the  `dept_id` is null.

@tbutterf When `dept_id` is `null` or `unaffiliated` or otherwise empty, we need to return the primary affiliation title and department values.

mlsamuelson (approved):
Behaves correctly when you test with the updated endpoint. My bad.

mlsamuelson:
Approved, but holding on merging until we have the SCHWEB endpoint scheduled for release.

See SCHWEB-1335

mlsamuelson:
Merging this now so we can get it into the SCHWEB update... that it also requires. :D Chicken and egg.  There will be a day or so delay before we get that subsequent update released. This was the last task for it.

PR #1306, opened 2024-06-26
fix(component-carousel): thumbnail navigation should only scroll when items are off screen

scott-williams-az:
Also fixed broken images just for this component
(this fix can be used going forward with other broken images)

### Description
The gallery thumbnails can suddenly swap position in the component when using them for navigation, as well as using the arrows. This issue is present in the current release, but the new max width makes it stand out a lot.

### Solution
Check full width of images and image container width before updating scroll position.
Do not scroll if full image width is less than image container width

### Links

- [JIRA ticket](https://asudev.jira.com/browse/WS2-2147)

### FOR APPROVERS


scott-williams-az:
This PR is close to fixing but I found an issue with there being 8 images and screen width 991px

mlsamuelson (approved):

PR #1311, opened 2024-07-03
fix(app-rfi): fix false error display on studentstatus

davidornelas11:
fix false error display on studentstatus field on submit.

## Description

The problem was that the payload was pointing to the same value in memory that got passed in and it was causing Formik to reset the state when the value was changed in the submission helpers. Creating a deep clone instead of a shallow clone fixes this


davidornelas11:
Closing PR for new PR

PR #1313, opened 2024-07-03
fix(app-rfi): fix default of student status error

davidornelas11:
### Description
fix default of student status error


mlsamuelson (approved):
Fixes both issues!

PR #1326, opened 2024-07-24
Hero H1 A11y issue

scott-williams-az:
### Description

This ticket has 3 items, This PR address item 1 `The doc-subtitle role is exposed as a heading with no level`
- The doc-subtitle role is meant for document, it is not meant for web

hgroup was being deprecated, however it may have made a come back as I still see relevant documentation:
- https://html.spec.whatwg.org/multipage/sections.html#the-hgroup-element
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup
There is another option to use header which does have an example of text before the h1
- https://html.spec.whatwg.org/multipage/sections.html#the-header-element

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1638) 

### FOR APPROVERS

### DIFF
--- a/packages/components-core/src/components/Hero/index.js
+++ b/packages/components-core/src/components/Hero/index.js
@@ -58,22 +58,27 @@ function headingHeroHtmlTemplate({
     undefined: "",
   };
 
-  return (
-    <div
-      className={classNames(`uds-hero`, {
-        [imageSize[image?.size]]: image?.size,
-      })}
-    >
-      <div className="hero-overlay"></div>
-      <HeroImage
-        className="hero"
-        src={image?.url}
-        alt={image?.altText}
-        data-testid="hero-image"
-      />
+  let titleElement = <></>;
 
-      {subTitle && (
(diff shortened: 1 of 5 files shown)


mlsamuelson (changes requested):
Looks good. Cleanly implemented. A commit for each package. Thx!

Because there’s a markup change in the `unity-bootstrap-theme` that will require consumers to update their markup, what do you think about having a [“breaking change” commit](https://github.com/semantic-release/semantic-release?tab=readme-ov-file#how-does-it-work) so it gets a breaking (major) release?  Historically that’s how we’ve handled it in the BS theme when markup changes. The React component doesn't need it because it handles its own markup.

scott-williams-az:
I thought about making it a breaking change, but because the code is still backwards compatible So that is why I did not, but I can make that commit.

mlsamuelson:
All good if it's backwards compatible. :) Thanks, @scott-williams-az .

mlsamuelson (approved):

PR #1331, opened 2024-07-25
UDS-1768: feat(component-header): removed display:unset on Header hidden inputs

juanmitriatti:
UDS-1768

feat(component-header): removed display:unset on Header hidden inputs

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1768)

### FOR APPROVERS

### DIFF
--- a/packages/component-header/src/components/UniversalNavbar/Search/index.styles.js
+++ b/packages/component-header/src/components/UniversalNavbar/Search/index.styles.js
@@ -9,7 +9,7 @@ const SearchWrapper = styled.form`
   .search-button {
     transition: cubic-bezier(0.19, 1, 0.19, 1);
   }
-  input {
+  input[name="q"] {
     width: 200px;
     background-color: #ffffff;
     border: 0;
@@ -20,6 +20,7 @@ const SearchWrapper = styled.form`
     color: #495057;
     border: 1px solid #707070;
     margin: 0.5rem 0;
+    display: unset;
     &:focus {
       border-color: transparent;
     }
--- a/packages/component-header/src/components/UniversalNavbar/Search/index.styles.js
+++ b/packages/component-header/src/components/UniversalNavbar/Search/index.styles.js
@@ -20,7 +20,6 @@ const SearchWrapper = styled.form`
     color: #495057;
     border: 1px solid #707070;
     margin: 0.5rem 0;
-    display: unset;
     &:focus {
       border-color: transparent;
     }


packages/component-header/src/components/UniversalNavbar/Search/index.styles.js
```diff
@@ -20,7 +20,6 @@ const SearchWrapper = styled.form`
     color: #495057;
     border: 1px solid #707070;
     margin: 0.5rem 0;
-    display: unset;
```
scott-williams-az: Looks good.
Can we also add some pattern to make sure it is only applying to search field?

Line 12 change `input` to:
`input[type="search"] ` or `input[name="q"]`

juanmitriatti: The inputs don't have the type, but I'll try adding them to see if it works.

scott-williams-az: The input should always have the name, so lets use `input[name="q"]` then

juanmitriatti: @scott-williams-az 
The problem are these inputs :
```
<input name="url_host" value={site} type="hidden" />
<input name="site" value="default_collection" type="hidden" />
<input name="sort" value="date:D:L:d1" type="hidden" />
<input name="output" value="xml_no_dtd" type="hidden" />
<input name="ie" value="UTF-8" type="hidden" />
<input name="oe" value="UTF-8" type="hidden" />
<input name="client" value="asu_frontend" type="hidden" />
<input name="proxystylesheet" value="asu_frontend" type="hidden" />
```
These were incorrectly read in Safari; the display: unset caused the tabbing through the site to pass through these input fields, which shouldn't happen.
So, the selector you suggested would not be enough : `input[name="q"]` .  I can use type=hidden as selector

scott-williams-az: Previously this code was targeting all the inputs, including the hidden inputs. 
```
input {
    width: 200px;
    ...
    display: unset;
```
This CSS rule is only meant for the one visible input, which is why I am suggesting we had `input[name="q"]`. We do not need to write any style rules for any hidden elements. By making this rule only apply to the visible input that also means we are not adding any styles (like a `display: unset`) to the hidden inputs.

juanmitriatti: @scott-williams-az  understood. Working on that fix.

juanmitriatti: Hi @scott-williams-az 
This is ready to test again.

mlsamuelson (approved):

PR #1332, opened 2024-07-29
Uds 1685: Add keyboard navigation to header

davidornelas11:
### Description
Added full keyboard navigation with header dropdown menus.

For testing:
1. Dropdown menus should be navigable with the keyboard arrows as well as tab
2. When inside a dropdown menu, the esc key should close the menu and put the focus at the parent link
3. Voiceover should recognize if the menu is expanded or not

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1685)

### FOR APPROVERS

### DIFF
--- a/packages/component-header/src/components/HeaderMain/NavbarContainer/DropdownItem/index.js
+++ b/packages/component-header/src/components/HeaderMain/NavbarContainer/DropdownItem/index.js
@@ -15,6 +15,9 @@ import { DropdownWrapper } from "./index.styles";
  *  items: [object][]
  *  buttons: Button[]
  *  classes?: string,
+ *  listId: string
+ *  setItemOpened: Function
+ *  parentLink: React.RefObject<HTMLElement> | null
  * }} DropdownItemProps
  */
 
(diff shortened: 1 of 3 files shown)


mlsamuelson (changes requested):
Couple questions, @davidornelas11. 

Because it's used for more than focusing the next link, what do you think about renaming `focusOnNextLink()` to `handleLinkFocus()` or something similar, to better reflect what that function does?

Do we need to implement similar updates for the Bootstrap version of the header?

davidornelas11:
> Couple questions, @davidornelas11.
> 
> Because it's used for more than focusing the next link, what do you think about renaming `focusOnNextLink()` to `handleLinkFocus()` or something similar, to better reflect what that function does?
> 
> Do we need to implement similar updates for the Bootstrap version of the header?

i agree on renaming the function. As far as the unity-bootstrap-theme, we would need to include the custom js file in the storybook with some instructions but it would just be for people who implement that instead of the react header.

mlsamuelson (changes requested):
Yes, so let's update the function name.

And thanks for the reminder that we're not currently shipping the header JS. It's not functionally equivalent in that regard at this time, so when we implement that, we can address sharing "current" JS. No work needed there, currently.

mlsamuelson (approved):

PR #1333, opened 2024-07-30
UDS-1773: feat(unity-bootstrap-theme): changed h4 to h2 and added styles for this change

juanmitriatti:
feat(unity-bootstrap-theme): changed h4 to h2 and added styles for this change

UDS-1773

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1773)

### FOR APPROVERS

### DIFF
--- a/packages/components-core/src/components/AnchorMenu/index.js
+++ b/packages/components-core/src/components/AnchorMenu/index.js
@@ -215,12 +215,12 @@ export const AnchorMenu = ({
               data-bs-target="#collapseAnchorMenu"
               aria-controls="collapseAnchorMenu"
             >
-              <h4>
+              <h2>
                 {menuTitle}:<i className="fas fa-chevron-down" />
-              </h4>
+              </h2>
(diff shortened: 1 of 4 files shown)


mlsamuelson (changes requested):
This looks good. I can see from the scss changes that this will be backwards compatible so this is good as-is.  In the future if you ever have a PR that updates a `unity-bootstrap-theme` component in a breaking way - usually it means the markup has changed and won't render correctly until templates are updated by people using the component - then you will want to make your commit message as a "breaking change."

There is also a version of the anchor menu in the `components-core` package, and this update should get made there, too. When we update one, we want to update the other. Can you add that to this PR? When you make the commit for those changes, you'll want to use `components-core` as the scope in the commit message so it gets an updated release.

mlsamuelson (approved):

mlsamuelson:
Thank you for the update. In both place, @juanmitriatti . Looks good.

PR #1335, opened 2024-07-31
UDS-1775: fix(unity-bootstrap-theme): switch anchor to button on accordions header and added styles

juanmitriatti:
fix(unity-bootstrap-theme): switch anchor to button on accordions header and added styles

UDS-1775

#### Solution
Switch over to use buttons instead of anchor tags ad that solves it for us without the need for custom javascript.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1775)

### FOR APPROVERS

### DIFF
--- a/packages/components-core/README.md
+++ b/packages/components-core/README.md
@@ -91,7 +91,7 @@ yarn test
 
 ```HTML
 <!-- Provide target divs for two carousels. Must have unique ids. -->
-<div id="default-card"></div>
+<div id="default-card"> </div>
 <div id="icon-card"></div>
 
 <!-- Include font awesome -->
--- a/packages/components-core/README.md
+++ b/packages/components-core/README.md
@@ -92,7 +92,7 @@ yarn test
 ```HTML
 <!-- Provide target divs for two carousels. Must have unique ids. -->
 <div id="default-card"> </div>
-<div id="icon-card"></div>
+<div id="icon-card"> </div>
 
 <!-- Include font awesome -->
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.3/css/all.min.css"
--- a/packages/components-core/src/components/Accordion/AccordionCard/index.js
+++ b/packages/components-core/src/components/Accordion/AccordionCard/index.js
@@ -24,12 +24,12 @@ export const AccordionCard = ({ id, item, openCard, onClick }) => (
   >
     <div className="accordion-header">
       <h4>
-        <a
+        <button
+          type="button"
           data-testid="accordion-opener"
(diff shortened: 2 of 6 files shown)


mlsamuelson (changes requested):
2 issues:

1. In testing this, I'm not seeing the focus state for the accordion header/button, and that's a regression. See the "states" tab for how that should be styeld: https://zeroheight.com/9f0b32a56/p/39a26f-accordion 

2. There is a React version of the accordion in `components-core` and it will need to get updated as well. The commit to add that update should use the `components-core` scope so it triggers a release for that package when this is merged. If you have questions about this version of the accordion, please reach out to Dave or Scott. https://asu.github.io/asu-unity-stack/@asu/components-core/index.html?path=/story/uds-accordion--default

juanmitriatti:
> 2 issues:
> 
> 1. In testing this, I'm not seeing the focus state for the accordion header/button, and that's a regression. See the "states" tab for how that should be styeld: https://zeroheight.com/9f0b32a56/p/39a26f-accordion
> 2. There is a React version of the accordion in `components-core` and it will need to get updated as well. The commit to add that update should use the `components-core` scope so it triggers a release for that package when this is merged. If you have questions about this version of the accordion, please reach out to Dave or Scott. https://asu.github.io/asu-unity-stack/@asu/components-core/index.html?path=/story/uds-accordion--default

Hi @mlsamuelson 
Regarding this first concern, please check these images : 

**Image 1** : This is what I locally see using the new approach "_w/ button: focus enabled in h4 or button_"
image
----------------------------------

**Image 2** : This is what I see using the previous approach "_w/ anchor, focus enabled in h4 tag_"
image

Would you please explain me the error that I'm introducing here? Thanks!

mlsamuelson:
Thanks for the question, @juanmitriatti . As we discovered when we Zoomed about this issue, it turns out that the missing focus state is a problem in the current `unity-bootstrap-theme` and `components-core` version of the accordions. It makes sense to fix it here. 

Please also check the Open and Close toggle on Accordions in multiple browsers. In my testing, the accordions only open in Chrome, but in Firefox they both open and close. The Firefox behavior is correct.

mlsamuelson (changes requested):
This PR updates both `unity-bootstrap-theme` and `components-core` but it only has commits for the `unity-boostrap-theme` scope. Can you do a commit with `fix(components-core)` as the scope so that component will get an updated release, please? You can do a small edit to the README.md in that component for that commit.

mlsamuelson:
Once we have a commit for the `components-core` scope, this should be good.

juanmitriatti:
Hi @mlsamuelson 
This is done. Please review the last commit.
Thanks!

mlsamuelson (approved):

PR #1337, opened 2024-08-07
fix(unity-bootstrap-theme): added css for long pagination on mobile

juanmitriatti:
UDS-1754

#### Solution
Added css for long pagination on mobile

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1754)

### FOR APPROVERS

### DIFF
--- a/packages/unity-bootstrap-theme/src/scss/extends/_pager.scss
+++ b/packages/unity-bootstrap-theme/src/scss/extends/_pager.scss
@@ -140,7 +140,9 @@ span.page-link {
 }
 @include media-breakpoint-down(md) {
   .pagination {
-    gap: .4rem;
+    gap: 0rem;
+    padding-right: 0.2rem;
+    padding-left: 0.2rem;
   }
   a.page-link {
     font-size: 14px;


mlsamuelson (approved):

scott-williams-az:
I was just looking at this, I am curious if we have our padding correct for the container class on mobile. 2rem seems high as the padding is .75rem on larger breakpoints. Sort of related to this ticket since the padding squeezes all the content. But this could be a separate ticket if the breakpoint padding needs adjusting

juanmitriatti:
> I was just looking at this, I am curious if we have our padding correct for the container class on mobile. 2rem seems high as the padding is .75rem on larger breakpoints. Sort of related to this ticket since the padding squeezes all the content. But this could be a separate ticket if the breakpoint padding needs adjusting

I added 'gap: 0' so that the pages don't get lost, in addition to the 'paddings' to center them. I would consult with @mlsamuelson  about whether to modify the 'container' in this ticket or create a new ticket, as I think it would be good to make the suggested change.

mlsamuelson:
@juanmitriatti I spoke with @scott-williams-az about this in standup the other day. The plan from that conversation was that he was going to create a separate ticket for the container padding issue. It does not need to be addressed in this PR.

scott-williams-az (approved):

PR #1348, opened 2024-08-28
feat(component-carousel): added aria-hidden for unvisible cards

juanmitriatti:
UDS-1805

#### Solution 
Added aria-hidden and tabindex for visible and unvisible cards

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1805)

### FOR APPROVERS

### DIFF
--- a/packages/component-carousel/src/core/components/BaseCarousel/glide/glide.setup.js
+++ b/packages/component-carousel/src/core/components/BaseCarousel/glide/glide.setup.js
@@ -163,6 +163,26 @@ function setNavButtonGradient(gliderElement, currentIndex, buttonCount) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


davidornelas11 (approved):
Looks good!

davidornelas11 (changes requested):
On further testing i realized that the buttons are still reachable via tabindex and voiceover. The parent element with a tabindex of -1 and aria-hidden true does not affect children that are inherently focusable.

Using the `inert` attribute seems to work for all children elements. It is available in all browsers as well
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert

mlsamuelson (approved):

davidornelas11 (approved):

PR #1354, opened 2024-09-06
fix(component-news): remove KE from default news header

Bob-E:
remove KE from default news header in src/core/constants/default-props.js

### Description

News component heading defaults to "Knowledge and enterprise news" and should be removed.

### Testing steps

In storybook, go to News component and remove header text in the controls. If the heading "Knowledge and enterprise news" doesn't appear, then the issue is fixed.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1811)


davidornelas11 (approved):
Looks good

PR #1358, opened 2024-09-09
Create script

scott-williams-az:
### Links

### FOR APPROVERS


packages/unity-react-core/package.json
```diff
@@ -28,6 +28,7 @@
     "url": "https://github.com/ASU/asu-unity-stack/issues"
   },
   "scripts": {
+    "create": "plop",
```
davidornelas11, line 31: Should we name this to be a bit more descriptive of what the script does?

davidornelas11 (changes requested):

packages/unity-react-core/package.json
scott-williams-az: Right now it is only creating components. So we could rename to `create-component` however, we could expand this to create hooks or other patterns. So I am open to naming.

davidornelas11: Another option would be to add a flag to read arguments so we could do "--component" for components or "--hook"? I also don't want to extend this PR just to add functionality so we can add these notes to the ticket for further brainstorming and I can merge this

scott-williams-az: plop already takes care of this if we add more options

PR #1366, opened 2024-09-20
fix(app-webdir-ui): prevent SearchResultsList from fetching titles when there are no results

davidornelas11:
prevent SearchResultsList from fetching titles when there are no results

Also added early return for any webdirectory that has the method of POST, since these already have the title info 

### FOR APPROVERS

Screen Shot 2024-09-20 at 11 59 41 AM

### DIFF
--- a/packages/app-webdir-ui/src/SearchResultsList/dataFormatter.js
+++ b/packages/app-webdir-ui/src/SearchResultsList/dataFormatter.js
@@ -100,9 +100,11 @@ export class ProfileService {
    * "title" and "dept_name" field to the individual results used in dataConverter.js
    */
   async processProfiles(term, filteredResults) {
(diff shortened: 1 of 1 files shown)


scott-williams-az:
Looks good. We do not have any coverage for if the API take too long or fails. Can we create a ticket for that, it might be too much to ask to include with this PR

scott-williams-az (approved):
Looks good. We do not have any coverage for if the API take too long or fails. Can we create a ticket for that, it might be too much to ask to include with this PR

davidornelas11:
> Looks good. We do not have any coverage for if the API take too long or fails. Can we create a ticket for that, it might be too much to ask to include with this PR

Agree on creating a ticket for this. It should encompass all the fetch requests so it may be a bit bigger than this ticket

PR #1368, opened 2024-09-24
chore(monorepo): Update git ignored files

scott-williams-az:
### Description

This will help while switching from branches based on current state and consolidated state


davidornelas11 (approved):
Approved

PR #1369, opened 2024-09-25
feat(app-rfi): only display plan if rfiDisplay true

scott-williams-az:
### Links

- [JIRA ticket](https://asudev.jira.com/browse/ERFI-159)

### FOR APPROVERS

### DIFF
--- a/packages/app-rfi/src/core/utils/datasource-helper.js
+++ b/packages/app-rfi/src/core/utils/datasource-helper.js
@@ -12,6 +12,7 @@
  * @prop {string[]} planCategories
  * @prop {[{strm: string, strmDescription: string}]} [applicationDeadlines]
  * @prop {string} title
+ * @prop {boolean} rfiDisplay
  */
 
 import { KEY } from "./constants";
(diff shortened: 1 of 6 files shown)


packages/app-rfi/src/core/utils/datasource-helper.js
```diff
@@ -106,6 +110,9 @@ export const normalizeDegreeData = data => {
 
         // Already formated title (BS)
         title: curr.title,
+
+        // Online Api does not have this field so we are going to always set it to true
+        rfiDisplay: true,
       };
```
mlsamuelson, line 116: Nice normalization pattern. Also appreciate the comment in this case. Thx!.

mlsamuelson (changes requested):
PR is missing handling for the following requirement:

> Update top level component display logic so if the component is preconfigured to display a plan that has rfiDisplay set to false, we display nothing. This would guard against status changes, forms configured for display false plans, AND for cases where Degree Listing pages send people to degrees that are set to rfiDisplay false.

mlsamuelson (approved):

PR #1370, opened 2024-09-27
Cc vitest sbname plop

scott-williams-az:
### Description
- autofix and rename stories (removing UDS, we can add parent folders again if we have a need to separate in the future)
- switch to vitest (Most tests pass, 4 tests do not and I think the tests need to be rewritten. Search the code for .skip to find them)
- prepare for new ts components `yarn create`
  - creates:
    - /src/components/NewComponent/NewComponent.tsx
    - /examples/new-component.html
    - /src/components/NewComponent/NewComponent.stories.tsx
    - /src/components/NewComponent/NewComponent.test.tsx
    - /src/components/NewComponent/index.js
  - adds (import/export statements):
    - /src/components/index.js
    - /src/core/utils/index.js
- added package script `yarn dev` starts storybook and vitest UI for auto feedback on testing

### Links

### FOR APPROVERS


packages/unity-react-core/examples/index.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
```
davidornelas11, line 1: Was this file made by mistake? Doesn't seem to have anything in it

packages/unity-react-core/package.json
```diff
@@ -29,8 +29,11 @@
   },
   "scripts": {
     "create": "plop",
+    "dev": "yarn storybook & yarn testui",
     "lint": "eslint --fix 'src/**/*.{js,jsx}' --ignore-path ../../.eslintignore",
-    "test": "jest --config=./jest.config.js --silent --coverage",
+    "jest": "jest --config=./jest.config.js --silent --coverage",
+    "test": "vitest",
```
davidornelas11: This is not blocking since watch false is automatically added in CI, but we should add `--watch=false` to the `test` script since we already have the testui script when we want to run it in watch mode with the ui when developing locally.

packages/unity-react-core/tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+  "compilerOptions": {
+    "moduleResolution": "bundler",
+    "allowImportingTsExtensions": true,
+    "emitDeclarationOnly":true,
+    "module": "ESNext",
+    "declaration": true,
+    "declarationDir": "dist/types",
+    "sourceMap": true,
+    "allowJs": true,
+    "rootDir": "src",
```
davidornelas11: There's a TS error here so we should either make `.` as the rootDir or remove `"**/*.ts",
    "**/*.tsx",
    "./vitest.setup.ts",` and add them to a "files" field.
Actually after further testing, there are errors all over the code base in my IDE since its reading the tsconfig file in unity-react-core and applying that everywhere. we might want to put this into it's own ticket and do it all at once unless we're all ok with the errors. It does make development a bit harder though

davidornelas11 (changes requested):
A few changes as well as possibly extracting the ts switchover into it's own branch and ticket

packages/unity-react-core/examples/index.html
scott-williams-az: I created this page so `http://localhost:5500/packages/unity-react-core/examples/` would have an index page with the navigation

scott-williams-az:
I think the alias "@shared" is the thing that is messing things up. so I might need to remove that for now

scott-williams-az:
I just removed the alias for now

PR #1373, opened 2024-10-02
chore: add preflight check to PR template

mlsamuelson:
### Description

Adding a preflight check section to the PR request template. Intent is that this section is reviewed and removed by the PR creator before the PR is submitted.  It's easy to forget to include tests or test updates and data layer additions and updates.

I'm hoping this won't be as easy to "look past" as the older checklist we had...

I didn't add it as a checklist to remain on the content... but we could do that. Thoughts?


davidornelas11:
I like it more as it is. Adding another checkbox might be too much

davidornelas11 (approved):

scott-williams-az (approved):

PR #1374, opened 2024-10-03
Erfi submit dupe prevention

mlsamuelson:
### Description

Formik's `isSubmitting` wasn't successfully disabling the RFI submit button when clicked. Lag time maybe? `submitCount` works better in testing.
git


mlsamuelson:
https://formik.org/docs/guides/form-submission

Another idea about why `isSubmitting` wasn't working out here could be that the value is set back to false faster than the component updates.  And the variability of all this could account for variations in what I was seeing when investigating the issue. To avoid timing issues, `submitCount` ought to be more effective.

packages/app-rfi/src/components/stepper/RfiStepperButtons.js
```diff
@@ -83,7 +83,7 @@ export const RfiStepperButtons = ({
               <Button
                 type="submit"
                 className="rfi-submit btn btn-primary"
-                disabled={!!isSubmitting}
+                disabled={submitCount > 0}
```
scott-williams-az: SubmitCount is increased each time the form tries to submit or go to the next page. One page variant click submit and the form will be stuck 
image

mlsamuelson: PR now uses the `totalSteps` in the disabled logic to account for steps getting logged as submits.

mlsamuelson: Needs more work: Paging back and forth through the form can exceed the submit count.

mlsamuelson:
Incorporated @scott-williams-az improvements that use useState() to make the submit disabling more reliable.

davidornelas11 (approved):
Tested in a webspark environment and it works as expected. The button is disabled and cannot be pressed again, even on keyboard.

PR #1381, opened 2024-10-16
lists

scott-williams-az:


packages/unity-react-core/examples/list.html
```diff
@@ -0,0 +1,140 @@
...
+
+  <!-- *************************************************************** -->
+  <!-- Load React. -->
+  <!-- Note: when deploying, replace "development.js" with "production.min.js". -->
+  <!-- ATTENTION! IF REACT IS NOT BUNDLED INTO THE LIBRARY UNCOMMENT
+        THOSE LINES BELOW
+      -->
+  <!-- *************************************************************** -->
+  <script src="https://unpkg.com/react@17/umd/react.production.min.js" crossorigin></script>
+  <script src="https://unpkg.com/react-dom@17/umd/react-dom.production.min.js" crossorigin></script>
```
mlsamuelson: Need to use React 18

mlsamuelson (changes requested):
Example not rendering, probably due to needing to use React 18.

Also, when list html rendes in storybook, when controls are set to default colors, the word `null` renders in the list's class attribute:
Screenshot 2024-11-21 at 4 19 41 PM

packages/unity-react-core/examples/list.html
scott-williams-az: done

mlsamuelson:
Screenshot and issue reported with the nulls was incorrect. Must have been looking at the wrong browser tab. Apologies.

mlsamuelson (approved):

mlsamuelson:
Didn't find a Jira ticket specifically for this story, but it's now merged and part of the larger effort, so I'm not going to bother creating a ticket at this point.

PR #1382, opened 2024-10-16
tooltip

scott-williams-az:


packages/unity-react-core/examples/tooltip.html
```diff
@@ -0,0 +1,61 @@
...
+
+  <!-- *************************************************************** -->
+  <!-- Load React. -->
+  <!-- Note: when deploying, replace "development.js" with "production.min.js". -->
+  <!-- ATTENTION! IF REACT IS NOT BUNDLED INTO THE LIBRARY UNCOMMENT
+        THOSE LINES BELOW
+      -->
+  <!-- *************************************************************** -->
+  <script src="https://unpkg.com/react@17/umd/react.production.min.js" crossorigin></script>
+  <script src="https://unpkg.com/react-dom@17/umd/react-dom.production.min.js" crossorigin></script>
```
mlsamuelson: Needs to use React 18.

packages/unity-react-core/src/components/Tooltip/index.js
mlsamuelson, line 1: We're using the filename `init.js` in the other components. Let's stick with that pattern.

mlsamuelson (changes requested):

mlsamuelson (approved):

PR #1384, opened 2024-10-16
divider

scott-williams-az:


packages/unity-react-core/examples/divider.html
```diff
@@ -0,0 +1,59 @@
...
+
+  <!-- *************************************************************** -->
+  <!-- Load React. -->
+  <!-- Note: when deploying, replace "development.js" with "production.min.js". -->
+  <!-- ATTENTION! IF REACT IS NOT BUNDLED INTO THE LIBRARY UNCOMMENT
+         THOSE LINES BELOW
+      -->
+  <!-- *************************************************************** -->
+  <script src="https://unpkg.com/react@17/umd/react.production.min.js" crossorigin></script>
+  <script src="https://unpkg.com/react-dom@17/umd/react-dom.production.min.js" crossorigin></script>
```
mlsamuelson: Using React 17 instead of 18.

packages/unity-react-core/src/components/Divider/index.js
mlsamuelson, line 1: Let's use init.js.

packages/unity-react-core/src/core/utils/index.js
```diff
@@ -131,6 +132,12 @@ export const initVideo = ({ targetSelector, props }) =>
 export const initTabbedPanels = ({ targetSelector, props }) =>
   RenderReact(TabbedPanels, props, document.querySelector(targetSelector));
 
+/**
+  * @param {ComponentProps} props
+  */
+export const initDivider = ({ targetSelector, props }) =>
```
mlsamuelson, line 138: I was going to say, Let's keep up the practice of putting these in alpha order... but I see we've already departed somewhere along the line. :P Not required. Consider it 5 bonus points. Get 1000 bonus points and you can redeem it for a stick of Juicy Fruit.

mlsamuelson (changes requested):

mlsamuelson (approved):

PR #1386, opened 2024-10-17
Uds 1857- update footer ranking logo

davidornelas11:
### Description
 update footer ranking logo for component-footer and unity-bootstrap-theme

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1857?atlOrigin=eyJpIjoiMWIxN2E4MTkwNjk1NDczMWEyZjFmODViMmZlYjMwYzUiLCJwIjoiaiJ9)


packages/unity-bootstrap-theme/stories/organisms/global-footer/global-footer.templates.js
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 
-import innovationLockup from "./200420-GlobalFooter-No1InnovationLockup.png";
+import innovationLockup from "./2024-ranked-1.png";
```
scott-williams-az: change
`import innovationLockup from "./2024-ranked-1.png";`
to
`import innovationLockup from "./../../../src/img/innovation-lockup/on-gold/2024-ranked-1.png";`
and delete
`packages/unity-bootstrap-theme/stories/organisms/global-footer/2024-ranked-1.png`

Even better if we can remove the duplication of `innovationLogo.png` and `2024-ranked-1.png` but I don't think we should complicate this ticket since those are in different packages. 

Approved with or without suggested change

scott-williams-az (dismissed):

mlsamuelson (changes requested):
@davidornelas11 Thoughts on Scott's requested change?

Also, what do you think about changing the file name from `2024-ranked-1.png` to `footer-rank.png` in order to be the most generic as possible so we don't need to rename the file every year and can simply do the swap-out by overwriting with the new file?

Pro: Only one action needed to do the update. Con: When it deploys out, caching may be harder to get cleared with CDNs if the filename is the same.

mlsamuelson (approved):

server/storybook-index-generator.js
```diff
@@ -203,7 +203,7 @@ module.exports = generateHTML = packages => `
                 href="https://www.asu.edu/rankings"
               >
                 <img
-                  src="./@asu/unity-bootstrap-theme/img/innovation-lockup/on-gold/200420-GlobalFooter-No1InnovationLockup.png"
+                  src="./@asu/unity-bootstrap-theme/img/innovation-lockup/on-gold/2024-ranked-1.png"
```
mlsamuelson, line 206: File needs to be renamed here, now, too.

PR #1389, opened 2024-10-24
Schweb 1356 - update profile result for people to fix number wrap

davidornelas11:
### Description
Update web directory contact line so number email do not wrap to next line. Address will still wrap
Web directory changes are dependent on the styles from unity-bootstrap-theme changes

### Links

- [JIRA ticket](https://asudev.jira.com/browse/SCHWEB-1356?atlOrigin=eyJpIjoiOWFkYzI2NGQyNDU2NDlkMGIzNmMyMGUwY2VlY2ZkMTUiLCJwIjoiaiJ9)

### DIFF
--- a/packages/app-webdir-ui/src/SearchPage/index.stories.js
+++ b/packages/app-webdir-ui/src/SearchPage/index.stories.js
@@ -12,7 +12,7 @@ export default {
 export const searchPageExample = () => (
   <div className="uds-content-align">
     <SearchPage
-      API_URL="https://test-asu-isearch.ws.asu.edu"
+      API_URL="https://search.asu.edu"
       searchApiVersion="/api/v1/"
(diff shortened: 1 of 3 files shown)


mlsamuelson (changes requested):
Progress! It does what was asked, but seeing it in action, there's more work needed.  See thread under https://asu-tech.slack.com/archives/C040QD2N4/p1729808012962529

mlsamuelson:
Main issue is - can we adjust to handle flex wrapping so we now don't end up with a one character column of address... Seems also that the container can get squished in some contexts and the wrap doesn't get triggered like normal.  Is the challenge tuning for breakpoints and flex wrapping?

mlsamuelson (approved):

PR #1392, opened 2024-10-30
feat(unity-bootstrap-theme): bundled all js together

davidornelas11:
### Description
bundled all js together

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1664?atlOrigin=eyJpIjoiNTNkMmY5MzdiZjJmNDFjMTgwNDg1NDAxMzc5N2E1MzQiLCJwIjoiaiJ9)

### DIFF
--- a/packages/unity-bootstrap-theme/src/js/data-layer.js
+++ b/packages/unity-bootstrap-theme/src/js/data-layer.js
@@ -1,154 +1,147 @@
-(function(){
-  function initDataLayer() {
-    /**
-     * Push events to data layer (Google Analytics)
-     * Used by Header and General events
-     * Required
-     */
-    const pushGAEvent = (event) => {
-      window.dataLayer = window.dataLayer || [];
-      const { dataLayer } = window;
-      if (dataLayer) dataLayer.push(event);
-    };
-
-    /**
-     * Header click events
-     * Elements with attribute [data-ga-header]
-     * 1 of 2 Header data layer functions
-     */
-    document.querySelectorAll('[data-ga-header]')?.forEach((element) =>
-      element.addEventListener('click', () => {
-        const event = element.getAttribute('data-ga-header-event') || 'link';
-        let action = element.getAttribute('data-ga-header-action') || 'click';
-        const expanded = element.getAttribute('aria-expanded');
-        if(expanded) {
-          action = expanded === 'false' ? 'open' : 'close';
-        }
-        const type = element.getAttribute('data-ga-header-type') || 'internal link';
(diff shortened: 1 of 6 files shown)


packages/unity-bootstrap-theme/src/js/index.js
```diff
@@ -0,0 +1,43 @@
+import "bootstrap/dist/js/bootstrap.esm";
+import {initDataLayer} from "./data-layer";
+import {initGlobalHeader} from "./global-header";
+import { initVideo } from "../../stories/atoms/video/video";
+import { initModals } from "../../stories/atoms/modals/modals";
+import { initializeFixedTable } from "../../stories/molecules/tables/tables";
+import { initBanner } from "../../stories/molecules/banners/banner";
+import { initCalendar } from "../..//stories/molecules/calendar/calendar";
```
mlsamuelson, line 8: Two slashes here by accident.

packages/unity-bootstrap-theme/src/js/data-layer.js
```diff
@@ -1,154 +1,147 @@
...
+      });
+    })
+  );
 
-  /* Function must be initialized after document load
-   * Example:
-   *   window.initDataLayer();
+  /**
+   * Input change event
+   * Input elementa with attribute [data-ga-input]
```
mlsamuelson, line 104: Typo: `elementa` instead of `elements`

packages/unity-bootstrap-theme/vite.config.js
```diff
@@ -52,15 +49,10 @@ const c = {
       treeshake: true,
       output: {
         entryFileNames: (chunkInfo) => {
-          if (chunkInfo.name.includes('index.esm')) {
-            return 'js/bootstrap.bundle.min.[format]';
-          }
-          return "js/[name].[format]";
+          return "js/unityBootstrap.bundle.js";
```
mlsamuelson, line 52: @davidornelas11 Would it require anything more than just reworking the output configs to allow for outputting the bundle as well as all the JS files individually? I'm thinking of situations where a consumer may be using `component-header` and so doesn't want to import that? The simple way to offer flexibility for cases like that would be to have individual as well as the bundle js files.

mlsamuelson: Also, for consistency, what do you think about using `-`s in the filename instead of camel case?

mlsamuelson (changes requested):

mlsamuelson:
Decided to cancel this as considering how js would be implemented downstream, it complicates rather than improves things.

@davidornelas11 to port renaming of output files from this PR to another PR.

PR #1397, opened 2024-11-05
Uds 1869 grid links

mlsamuelson:
### Description

Adds Grid Links

Update: Data layer now included. Includes code other components can leverage.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1869)

### DIFF
--- a/packages/unity-react-core/src/components/GridLinks/GridLinks.stories.tsx
+++ b/packages/unity-react-core/src/components/GridLinks/GridLinks.stories.tsx
@@ -4,13 +4,27 @@ import { GridLinks, GridLinksProps } from "./GridLinks";
 import {
   gridLinksNumColumns,
   gridLinksTextColor,
-  gridLinksBgColor,
-  gridLinksBgColorClassName,
 } from "./GridLinksConstants";
+
+const gridLinksBgColorClassName = {
+  None: "",
+  "Gray 1": "bg-gray-1",
+  "Gray 7": "bg-gray-7",
+};
 
 export default {
   title: "Components/Grid Links",
   component: GridLinks,
+  argTypes: {
+    bgColor: {
+      name: "Background color of the section",
+      description: "Background color is not applied to the component, but to the section wrapper.",
+      options: Object.keys(gridLinksBgColorClassName),
+      control: {
+        type: "radio",
+      },
+    }
+  }
 };
 
 const gridLinkItems = [
@@ -61,8 +75,6 @@ const defaultStoryProps = {
   textColor: gridLinksTextColor.NONE,
   gridLinkItems,
   // children: "Content to be displayed",
-  // bgColor is used story section wrapper, not in component.
-  bgColor: gridLinksBgColor.NONE,
   // Display the data-ga attributes on the links for Storybook HTML references.
   useExternal: true,
 };
--- a/packages/unity-react-core/src/components/GridLinks/GridLinks.stories.tsx
+++ b/packages/unity-react-core/src/components/GridLinks/GridLinks.stories.tsx
@@ -1,0 +1,82 @@
+import React from "react";
+
+import { GridLinks, GridLinksProps } from "./GridLinks";
+import {
+  gridLinksNumColumns,
+  gridLinksTextColor,
+  gridLinksBgColor,
+  gridLinksBgColorClassName,
+} from "./GridLinksConstants";
+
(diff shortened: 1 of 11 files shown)


packages/unity-react-core/src/components/GridLinks/GridLinksConstants.ts
```diff
@@ -0,0 +1,28 @@
...
+  TEXT_GOLD = "gold",
+  TEXT_WHITE = "white",
+}
+export const gridLinksTextColorClassName = {
+  [gridLinksTextColor.NONE]: "",
+  [gridLinksTextColor.TEXT_GOLD]: "text-gold",
+  [gridLinksTextColor.TEXT_WHITE]: "text-white",
+};
+
+export enum gridLinksBgColor {
```
scott-williams-az: What do you think about removing from Constants as this is only needed for storybook. I will leave a comment in that file for what it could look like

packages/unity-react-core/src/components/GridLinks/GridLinks.stories.tsx
```diff
@@ -0,0 +1,82 @@
+import React from "react";
+
+import { GridLinks, GridLinksProps } from "./GridLinks";
+import {
+  gridLinksNumColumns,
+  gridLinksTextColor,
+  gridLinksBgColor,
+  gridLinksBgColorClassName,
+} from "./GridLinksConstants";
+
```
scott-williams-az, line 8: See note in Constants file about removing `  gridLinksBgColor`, `gridLinksBgColorClassName`

Since this is not a prop on the component we need to manually set the controls

```
const gridLinksBgColorClassName = {
  None: "",
  "Gray 1": "bg-gray-1",
  "Gray 7": "bg-gray-7",
};

export default {
  title: "Components/Grid Links",
  component: GridLinks,
  argTypes: {
    bgColor: {
      name: "Background Color of the section",
      description: "Background color is not applied to the component, but to the section wrapper.",
      options: Object.keys(gridLinksBgColorClassName),
      mapping: gridLinksBgColorClassName,
      control: {
        type: "radio",
      },
    },
  },
};
```

packages/unity-react-core/src/components/GridLinks/GridLinks.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+  bgColor: gridLinksBgColor.NONE,
+};
+
+const defaultBgColor = gridLinksBgColor.NONE;
+
+const gridLinksTemplate = ({ bgColor, ...args }) => (
+  <section className={gridLinksBgColorClassName[bgColor]}>
+    <GridLinks {...(args as GridLinksProps)} />
+  </section>
+);
```
scott-williams-az:
```
const defaultStoryProps = {
  numColumns: gridLinksNumColumns.TWO_COLUMNS,
  textColor: gridLinksTextColor.NONE,
  gridLinkItems: gridLinkItems,
  // children: "Content to be displayed",
};

const gridLinksTemplate = ({ bgColor = gridLinksBgColorClassName.None, ...args }) => (
  <section className={bgColor}>
    <GridLinks {...(args as GridLinksProps)} />
  </section>
);
```

shared/utils/datalayer-render.js
```diff
@@ -0,0 +1,21 @@
+// @ts-check
+import { trackGAEvent } from "../services/googleAnalytics";
+
+const convertToDataAttr = d => ({
```
davidornelas11, line 4: I know we add the data when we use the dataLayerRender function in a component, but should we also do some defaults here? If any of these aren't supplied they will render as "undefined". From an analytics standpoint, I'm not sure is this is ok or not

davidornelas11 (approved):
Approved with a comment

shared/utils/datalayer-render.js
mlsamuelson: Our thinking was that it's the job of the component itself to provide the defaults. See https://github.com/ASU/asu-unity-stack/pull/1397/files#diff-cd97d9900a081811f3e0bf92ff3f90bad957136df1d0b2c3f03e104ef323bad5R10-R16

davidornelas11: Yea I saw that, I just wanted to make sure. Everything looks good then

scott-williams-az (approved):

PR #1402, opened 2024-11-07
fix(unity-react-core): add data layer to grid links

mlsamuelson:
### Description

We needed a pattern for the data layer that could be used for this and other instances of design components that are implemented in `unity-react-core`. Requirements
* Allows for "non-React" output that shows the `data-ga*` attributes like we use in CMS contexts
* Allows for default React behavior that uses existing internal datalayer handling and doesn't display `data-ga*` attributes.
* Allows for configurability and customization to for component data layer object variations.
* Displays helpful and meaningful example output and controls within the Storybook UI.

Adds datalayer-render.js shared util.

This PR was made separate from the rest of the ticket so it can be used as a reference to the proposed pattern.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1869)

### DIFF
--- a/shared/utils/datalayer-render.js
+++ b/shared/utils/datalayer-render.js
@@ -1,0 +1,23 @@
+// @ts-check
+import { trackGAEvent } from "../services/googleAnalytics";
+
+export function dataLayerRender(gaObject, useExternal) {
+  if (useExternal) {
+    return {
+      "data-ga": gaObject.text,
+      "data-ga-name": gaObject.name,
+      "data-ga-event": gaObject.event,
+      "data-ga-action": gaObject.action,
+      "data-ga-type": gaObject.type,
+      "data-ga-region": gaObject.region,
+      "data-ga-section": gaObject.section,
+      "data-ga-component": "",
+    }
+  } else {
+    return {
+      onClick: (e) => {
+        trackGAEvent(gaObject);
+      },
+    };
+  }
+}
--- a/shared/utils/datalayer-render.js
+++ b/shared/utils/datalayer-render.js
@@ -1,23 +1,21 @@
 // @ts-check
 import { trackGAEvent } from "../services/googleAnalytics";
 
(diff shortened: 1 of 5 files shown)


shared/utils/datalayer-render.js
```diff
@@ -0,0 +1,23 @@
...
+      "data-ga-component": "",
+    }
+  } else {
+    return {
+      onClick: (e) => {
+        trackGAEvent(gaObject);
+      },
+    };
+  }
+}
```
scott-williams-az, line 21:
```
const convertToDataAttr = d => ({
      "data-ga": d.text,
      "data-ga-name": d.name,
      "data-ga-event": d.event,
      "data-ga-action": d.action,
      "data-ga-type": d.type,
      "data-ga-region": d.region,
      "data-ga-section": d.section,
      "data-ga-component": d.component,
    });

const convertToClickHandler = d => ({ onClick: (e) => useGAEvent(d)});

export function dataLayerRender(gaObject, useExternal) {
  return useExternal ?
    convertToDataAttr(gaObject) :
    convertToClickHandler(gaObject);
}
```

Maybe we can still split things out into 2 internal functions and the one exported function

do we need fallbacks? ex: `"data-ga-component": d.component || ""`,

shared/utils/datalayer-render.js
```diff
@@ -0,0 +1,23 @@
...
+  if (useExternal) {
+    return {
+      "data-ga": gaObject.text,
+      "data-ga-name": gaObject.name,
+      "data-ga-event": gaObject.event,
+      "data-ga-action": gaObject.action,
+      "data-ga-type": gaObject.type,
+      "data-ga-region": gaObject.region,
+      "data-ga-section": gaObject.section,
+      "data-ga-component": "",
```
scott-williams-az: probably not always blank

scott-williams-az (approved):

PR #1417, opened 2024-12-06
Ga event wrapper

scott-williams-az:
### Links

### FOR APPROVERS

### DIFF
--- a/packages/unity-react-core/.storybook/decorators.tsx
+++ b/packages/unity-react-core/.storybook/decorators.tsx
@@ -2,7 +2,7 @@
  * This file houses all non-addon related decorators
  */
 import { Decorator } from '@storybook/react';
-import React, { useEffect, forwardRef, useLayoutEffect, StrictMode } from 'react';
+import React, { useLayoutEffect, StrictMode } from 'react';
 import { renderToStaticMarkup } from 'react-dom/server';
 
 import { identifierPrefix } from '../src/components/GaEventWrapper/useBaseSpecificFramework';
@@ -28,9 +28,10 @@ const StaticStory = ({args, Container, children, rootRef}) => {
     /**
      * Storybook only useId() will prefix the id with this identifier allowing
      * us to identify when the output is meant for bootstrap (non react)
-     */
+    */
     rootRef.current.innerHTML = renderToStaticMarkup(children, { identifierPrefix: identifierPrefix });
 
+    window.initDataLayer()
   }, [args]);
   return <Container rootRef={rootRef}></Container>;
 }
--- a/packages/unity-react-core/.storybook/preview.js
+++ b/packages/unity-react-core/.storybook/preview.js
@@ -2,6 +2,7 @@ import "@asu/unity-bootstrap-theme/src/scss/unity-bootstrap-theme.bundle.scss";
 import "../../../node_modules/bootstrap/dist/js/bootstrap.bundle.min.js";
 import { globalDecorators } from "./decorators.tsx";
 
+import "@asu/unity-bootstrap-theme/src/js/data-layer.js";
 const parameters = {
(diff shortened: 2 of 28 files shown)


packages/unity-react-core/src/components/RankingCard/RankingCard.jsx
scott-williams-az, line 1: Previously this code required some JS in order to function, as seen here [ranking-card-large](https://asu.github.io/asu-unity-stack/@asu/unity-bootstrap-theme/index.html?path=/story/molecules-cards-examples--ranking-card-large)

```
function rankingFunc() {
  const $infoLayer = document.querySelector(".info-layer"),
    $toggleIcon = document.getElementById("dispatch");
  $toggleIcon?.addEventListener("click", function () {
    $infoLayer?.classList.toggle("active");
  });
}
```
With just a few changes we can leverage the normal Bootstrap functions to allow opening and closing. This file and also [_ranking-cards.scss](https://github.com/ASU/asu-unity-stack/pull/1417/files#diff-339d8f36b1c5815a7e88c0da3d5f4adcc07187e13d56a610457e701d2877d6e9R76-L76)

The small variant works perfectly.

There internal components in this file that were used one time, I just moved those to be inline.
The large variant works, but the opening effect is janky.

packages/unity-react-core/.storybook/decorators.tsx
```diff
@@ -28,9 +28,10 @@ const StaticStory = ({args, Container, children, rootRef}) => {
     /**
      * Storybook only useId() will prefix the id with this identifier allowing
      * us to identify when the output is meant for bootstrap (non react)
-     */
+    */
     rootRef.current.innerHTML = renderToStaticMarkup(children, { identifierPrefix: identifierPrefix });
 
+    window.initDataLayer()
```
scott-williams-az, line 34: This is inside the bootstrap decorator. This is how we log events with bootstrap so we need to initialize here.

packages/unity-react-core/.storybook/preview.js
```diff
@@ -2,6 +2,7 @@ import "@asu/unity-bootstrap-theme/src/scss/unity-bootstrap-theme.bundle.scss";
 import "../../../node_modules/bootstrap/dist/js/bootstrap.bundle.min.js";
 import { globalDecorators } from "./decorators.tsx";
 
+import "@asu/unity-bootstrap-theme/src/js/data-layer.js";
```
scott-williams-az, line 5: technically loads on all stories, but we only run it with bootstrap decorator

packages/unity-react-core/src/components/RankingCard/RankingCard.jsx
```diff
@@ -27,120 +27,87 @@ const AVAILABLE_SIZES = {
...
-  heading: PropTypes.string.isRequired,
-  citation: PropTypes.string.isRequired,
-};
-
 const InfoLayerWrapper = ({ imageSize, body, heading, readMoreLink }) => {
   const [open, setOpen] = useState(false);
   // TODO: Switch to useId when we upgrade to React 18
-  const uniqueId = `info-layer-${Math.floor(Math.random() * 100000)}`;
-
+  const { id, isReact, isBootstrap } = useBaseSpecificFramework();
```
davidornelas11: I noticed the useBaseSpecificFramework hook doesn't return any id in it's return statement so where would this ID come from?

scott-williams-az: I went back and forth on if I should export id, I think even though it uses useId and could return it, we will keep the useBaseSpecificFramework hook focused on one thing and just call useId as needed (open to thoughts)

mlsamuelson (approved):

PR #1421, opened 2024-12-12
fix(unity-react-core): add Modal story

davidornelas11:
### Description

add Modal story

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1864)


packages/unity-react-core/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,55 @@
+import React, {useEffect} from "react";
+/**
+ *
+ * TODO: Should we be using bootstrap's built in modal functionality?
```
scott-williams-az, line 6: If this is this possible to use bootstrap I'd like to see if that works.

packages/unity-react-core/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+      document
+        ?.getElementById('closeModalButton')
+        .addEventListener('click', function () {
+          document.getElementById('uds-modal').classList.remove('open');
+        });
+  }, []);
+
+  return (
+    <div className="container-fluid">
+      <button id="openModalButton" className="btn btn-dark">
```
scott-williams-az: I know this is example code only so I would like to capture the direction we want to move in. Can you add these notes as comments (if you agree) to the component?
 - I'd like the Modal to only be the content of the modal. (without the button).
 - React 18 has createPortal. Maybe we use this so the modal can be attached to the bottom of the body?
 - We should be able to use the default button component to trigger the modal,
 - We should also be able to trigger the modal on other events,

packages/unity-react-core/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+        Show modal
+      </button>
+
+      <div id="uds-modal" className="uds-modal">
+        <div className="uds-modal-container">
+          <button
```
scott-williams-az: Can you swap this out and use `ButtonIconOnly`

packages/unity-react-core/examples/js/example-helper.js
```diff
@@ -7,6 +7,7 @@ const toolBar = `
           <div class="uds-quicklinks uds-quicklinks-expanded-md">
             <div id="collapseExample" class="card">
               <nav class="nav" aria-label="Same Page">
+                <a class="nav-link" href="modal.html"><span class="fas fa-hand-spock"></span>Modal</a>
```
scott-williams-az: remove

packages/unity-react-core/examples/modal.html
```diff
@@ -0,0 +1,53 @@
+<!DOCTYPE html>
```
scott-williams-az: delete file

packages/unity-react-core/src/components/Modal/index.js
```diff
@@ -0,0 +1 @@
+export { initModal as default } from "../../core/utils";
```
scott-williams-az: delete file or comment line file should also be named init.js (not your fault, my fault with the plop generator)

packages/unity-react-core/src/core/utils/index.js
```diff
@@ -21,6 +21,7 @@ import { GridLinks } from "../../components/GridLinks/GridLinks.tsx";
 import { Hero } from "../../components/Hero/Hero";
 import { Image } from "../../components/Image/Image";
 import { List } from "../../components/List/List.tsx";
+import { Modal } from "../../components/Modal/Modal.tsx";
```
scott-williams-az: I think this is only needed if we export, we have already given ourself a reference to what we are not exporting in src/components/index.js so lets avoid merge conflicts and not update this file unless the component is being exported

packages/unity-react-core/src/components/Modal/Modal.tsx
davidornelas11: This is gonna require a bit of a rewrite so I can make a ticket but I say for now we just leave it as is since we don't even really use it anywhere

davidornelas11:
Ready for re review

packages/unity-react-core/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,79 @@
...
+          <GaEventWrapper
+            gaData={{...defaultGaData, ...gaData}}
+            >
+          <ButtonIconOnly
+          // @ts-ignore
+            id="closeModalButton"
+            className="uds-modal-close-btn"
+            icon={['fas', 'times']}
+          >
+          </ButtonIconOnly>
```
scott-williams-az: `<ButtonIconOnly .... ></ButtonIconOnly>` -> `<ButtonIconOnly .... />`

davidornelas11: Good catch. Just fixed

PR #1422, opened 2024-12-12
Uds 1866 notification banner

scott-williams-az:
Some work with GaEventWrapper
Adding Notification banner as example only (We can export it, as it might be good)

I am not really concerned about where these are added at this time, but we will want to follow up and figure this out before we are done `title: "More Examples/Notification Banner",`

I have added notes to the doc page, but also created this follow up ticket as a nice to have: (Unity-react-core: Storybook Custom doc generator)
https://asudev.jira.com/browse/UDS-1913

### DIFF
--- a/packages/unity-react-core/src/components/NotificationBanner/NotificationBanner.stories.tsx
+++ b/packages/unity-react-core/src/components/NotificationBanner/NotificationBanner.stories.tsx
@@ -1,0 +1,51 @@
+import React from "react";
+
+import {
+  NotificationBanner,
+  NotificationBannerProps,
+} from "./NotificationBanner";
+
+export default {
+  title: "More Examples/Notification Banner",
+  component: NotificationBanner,
+  args: { color: "orange" },
+  parameters: {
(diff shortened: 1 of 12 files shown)


packages/unity-react-core/src/components/NotificationBanner/NotificationBanner.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+    },
+  ],
+};
+
+const notificationBannerTemplate = args => <NotificationBanner {...args} />;
+
+export const Overview = {
+  render: notificationBannerTemplate.bind({}),
+  args: {
+    ...defaultProps,
```
davidornelas11, line 56: Should we remove the children prop from the args here for all stories that have it? It results in an arg control that may be confusing to people coming to storybook
Screenshot 2024-12-17 at 4 46 49 PM.
Other than that, it looks good

davidornelas11 (approved):
Approved with suggestions but not required

PR #1430, opened 2024-12-19
chore(unity-react-core): add Table story

davidornelas11:
### Description
add Table story

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1878?atlOrigin=eyJpIjoiNTFhMWMzMjA5YjhmNGUzYzlkMDgyYWEzM2ZmZmI3NGYiLCJwIjoiaiJ9)


mlsamuelson (dismissed):
This is missing the fixed table variation. We do need to retain the reference for that.

mlsamuelson (changes requested):
If I switch to the MD mobile view on the fixed view table, the gradient at the edge of the horizontally scrollable area is off. Is the JS to initialize the fixed area not included or executing? In the pre-consolidation story we have the Javascript tab, which shows this.

Path to the JS file `packages/unity-bootstrap-theme/stories/molecules/tables/tables.js`

mlsamuelson (commented):
Just want to confirm on this that the fix is applied to the React version - looks good. But the Bootstrap version will be like the other stories that require JS, where we'll want to default to the React view.

Also, is the error on the Tables Docs page known and to be resolved elsewhere?

mlsamuelson (approved):

PR #1439, opened 2025-01-09
feat(unity-react-core): form elements

scott-williams-az:


mlsamuelson (approved):
Only issues are non-blocking - displaySize class doesn't do anything on the Select component - but I don't know that I needs to. Can be adjusted later if needed. And then there's no prop for doing select lists with multiples lines visible. But again, that's one we can enable if/when it's needed.

Approving.

PR #1447, opened 2025-01-16
Update header to include fontawesome and fix navbar scrolling issue

davidornelas11:
Update header to include fontawesome and fix navbar scrolling issue


scott-williams-az (approved):

scott-williams-az:
Looks good except for the lock file (needs to be regenerated)

mlsamuelson (approved):

PR #1448, opened 2025-01-17
Fix profile squishing grid view

davidornelas11:
Fix profile squishing grid view
### Links

- [JIRA ticket](https://asudev.jira.com/browse/WS2-2190)

### FOR APPROVERS


mlsamuelson (approved):
2 minor things - the control is not on the faculty rank story (it’s in a different location, so probably missed), also, once selected, there’s no way to reverse it without reloading the page. … I’m okay with moving forward here, though.

PR #1473, opened 2025-02-13
uds-1870: add deprecation warnings to newly deprecated packages

scott-williams-az:
### Description
This assumes we should leave newly deprecated packages in the repo for a given time, incase we need to provided limited support

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1870)

### DIFF
--- a/packages/component-footer/package.json
+++ b/packages/component-footer/package.json
@@ -8,7 +8,7 @@
   "types": "./dist/main.d.ts",
   "author": "Conner Shannon",
   "license": "MIT",
-  "deprecated": "This package has been moved. Please use that @asu/component-header-footer to receive continued support.",
+  "deprecated": "This package has been moved. Please use @asu/component-header-footer to receive continued support.",
   "files": [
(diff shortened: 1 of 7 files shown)


packages/component-footer/package.json
```diff
@@ -8,6 +8,7 @@
   "types": "./dist/main.d.ts",
   "author": "Conner Shannon",
   "license": "MIT",
+  "deprecated": "This package has been moved. Please use that @asu/component-header-footer to receive continued support.",
```
mlsamuelson: @scott-williams-az Check all the "deprecated" text update lines like this one... `Please use that @asu/...`  the `that` needs to be removed from them all.

mlsamuelson (changes requested):

packages/component-footer/package.json
scott-williams-az: fixed typo (removed `that`)

mlsamuelson (approved):

PR #1476, opened 2025-02-17
fix(unity-react-core): update AnchorMenu Ex/Doc

scott-williams-az:
### Description

This request is from Sarah Hough:  Please update the Title Case - all nav is in title case.
Please update the title case in the AnchorMenu for all containers to match what is in the UDS Components.

### Solution

We do need to enforce this standard and have added clear usage requirements. We can not enforce this with code as a code solution can not be achieved with 100% accuracy.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1762)

### DIFF
--- a/packages/unity-react-core/src/components/AnchorMenu/AnchorMenu.stories.jsx
+++ b/packages/unity-react-core/src/components/AnchorMenu/AnchorMenu.stories.jsx
@@ -2,225 +2,82 @@
 import classNames from "classnames";
 import React from "react";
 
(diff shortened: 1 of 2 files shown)


mlsamuelson (approved):

scott-williams-az:
Bootstrap version of Anchor Menu is not working. I have disabled the story from loading, but this needs to be fixed. Created ticket [UDS-1958](https://asudev.jira.com/browse/UDS-1958)

PR #1487, opened 2025-03-05
build: stash build artifacts to reduce build time

davidornelas11:
stash build artifacts to reduce build time


davidornelas11:
CLosing this pr as the time saved on stashing node_modules files takes roughly the same amount of time as running yarn install every time. In the future, we may save time by stashing the build files and only running yarn install instead of yarn install and yarn build

PR #1493, opened 2025-03-13
UDS-1975: fix(component-events): added style to fix a:focus

juanmitriatti:
UDS-1975

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1975)

### DIFF
--- a/packages/unity-bootstrap-theme/README.md
+++ b/packages/unity-bootstrap-theme/README.md
@@ -1,3 +1,4 @@
 # `@asu/unity-bootstrap-theme`
 
 Please see the [Getting Started page for the Unity Bootstrap Theme](https://asu.github.io/asu-unity-stack/@asu/unity-bootstrap-theme/index.html)
+
--- a/packages/unity-bootstrap-theme/src/scss/extends/_misc.scss
+++ b/packages/unity-bootstrap-theme/src/scss/extends/_misc.scss
@@ -66,6 +66,8 @@ label {
 .card-title a {
   color: $uds-color-base-gray-7;
   text-decoration: none;
+  display: inline-block;
+  margin: 4px;
 }
 .card-title a:hover {
   text-decoration: underline;
@@ -126,6 +128,5 @@ textarea:focus,
 select:focus {
   outline: none !important;
   box-shadow: 0px 0px 0px 2px $uds-color-base-white, 0px 0px 0px 4px $uds-color-base-gray-7 !important;
-  display: inline-block;
-  margin: 4px;
+  z-index: 1;
 }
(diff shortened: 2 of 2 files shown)


mlsamuelson (dismissed):
Fix is causes layout shift when the images and links are focused. Are we able to use the same technique as with other focus instances so we avoid that?

juanmitriatti:
> Fix is causes layout shift when the images and links are focused. Are we able to use the same technique as with other focus instances so we avoid that?

I added `margin: 4px` to make the `box-shadow` visible on focus. Without this margin, the wrapper tag (for example, in the `h3` of the `title`) hides it, causing the 'movement' effect when focusing. Also, `display: inline-block` ensures that the box-shadow applies globally rather than per line.

With display: inline-block
image

Without
image

davidornelas11:
@mlsamuelson The overflow hidden on the parent h3 causes the shadow to not appear. The hidden is there because the heading shouldn't pass 3 lines so that's necessary

mlsamuelson:
Is there any way we can accomplish this without the layout shifting when focus occurs? What are our options?

If we can't change it, we will need to have The Brand team review the behavior for acceptability.

juanmitriatti:
> Is there any way we can accomplish this without the layout shifting when focus occurs? What are our options?
> 
> If we can't change it, we will need to have The Brand team review the behavior for acceptability.

The box-shadow on focus extends beyond the element’s box, but since overflow: hidden is applied, it gets clipped. We can’t make it visible without adding a margin, and adding it causes the layout shift.

juanmitriatti:
Hi @mlsamuelson , @davidornelas11 
I found a possible solution: add `margin: 4px` and `display: inline-block` to the a .card-title styles to ensure the box shadow is visible, and remove these styles from a:focus.
The problems/concerns that we've here are :

- Image left border does not show up (it's because the wrapper class has `overflow: hidden`) . (Please check first image attached)
 In order to fix this, we could add `margin: 0 4px` to the wrapper ul classes. (Please check second image attached).
 
- On the other hand, all the titles will have margin: 4px.

1) **First image:**
WhatsApp Image 2025-03-20 at 10 30 54

2) **Second image:**

WhatsApp Image 2025-03-20 at 10 45 27

mlsamuelson:
I'm less concerned with the images, and more concerned with the text - that should be preventable. If we can avoid the text from shifting, then I'm okay with allowing the images to shift some.

mlsamuelson:
Conversation on this [here](https://asu-tech.slack.com/archives/CT01Z74KY/p1743107153407559?thread_ts=1743103335.772769&cid=CT01Z74KY)

davidornelas11:
This has been tested and double checked by Alysia on the brand team. Looks good. Will approve and merge

davidornelas11 (dismissed):

davidornelas11 (approved):

PR #1494, opened 2025-03-14
fix(component-header-footer): pass onClick handler to header buttons

davidornelas11:
### Description
Fixed an issue where the onClick function prop wasn't being properly passed down to buttons in the header component, while all other props were correctly passing through. Also added tests to check for this.

Added console errors when somebody passes an href and an on onCLick to a button since that breaks accessibility. Does not crash the component but it should help in case any accessibility issues arise

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1976)

### DIFF
--- a/packages/component-header-footer/src/header/components/Button/index.js
+++ b/packages/component-header-footer/src/header/components/Button/index.js
@@ -20,6 +20,7 @@ const Button = ({ href, color, text, classes, onClick, onFocus }) => {
       className={`button-${color} ${classes ?? ""}`}
       onClick={onClick}
       onFocus={onFocus}
+      {...(onClick && { role: "button" })}
     >
       {text}
     </ButtonWrapper>
--- a/packages/component-header-footer/src/header/components/HeaderMain/NavbarContainer/index.js
+++ b/packages/component-header-footer/src/header/components/HeaderMain/NavbarContainer/index.js
@@ -10,6 +10,9 @@ import { UniversalNavbar } from "../../UniversalNavbar";
 import { Wrapper } from "./index.styles";
 import { NavItem } from "./NavItem";
 
+export const BUTTON_ERROR_MESSAGE =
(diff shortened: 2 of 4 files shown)


packages/component-header-footer/src/header/components/Button/index.js
```diff
@@ -20,6 +20,7 @@ const Button = ({ href, color, text, classes, onClick, onFocus }) => {
       className={`button-${color} ${classes ?? ""}`}
       onClick={onClick}
       onFocus={onFocus}
+      {...(onClick && { role: "button" })}
```
mlsamuelson, line 23: Does the onClick on line 21 still need to be there now?

packages/component-header-footer/src/header/components/HeaderMain/NavbarContainer/index.js
```diff
@@ -19,6 +22,26 @@ const NavbarContainer = () => {
     setItemOpened(prev => (itemOpened === itemId ? undefined : itemId));
   };
 
+  const validateButton = button => {
+    if (button.onClick && button.href) {
+      console.error(BUTTON_ERROR_MESSAGE);
+    }
+  };
```
mlsamuelson, line 29: Thank you for adding this. Love the forward-thinking here.

mlsamuelson (changes requested):

packages/component-header-footer/src/header/components/Button/index.js
davidornelas11: Yes this just checks if it’s there and adds the proper role for accessibility. It doesn’t invoke the function

mlsamuelson: Ah, alright. I spaced on the `&&` there. ;)

mlsamuelson (approved):

PR #1513, opened 2025-04-09
UDS-1965: fix(app-rfi): added autofocus property for Campus select

juanmitriatti:
## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1965)

### DIFF
--- a/packages/app-rfi/src/components/appVariants/index.js
+++ b/packages/app-rfi/src/components/appVariants/index.js
@@ -83,7 +83,7 @@ const variants = {
         To learn more about ASU or a specific program, fill out the form below
         then check your email.
       </p>
-      <Campus autoFocus gaData={gaData1Of2} />
+      <Campus gaData={gaData1Of2} />
       <CareerAndStudentType gaData={gaData1Of2} />
       <Interest1 gaData={gaData1Of2} />
       <Interest2 gaData={gaData1Of2} />
--- a/packages/app-rfi/src/components/appVariants/index.js
+++ b/packages/app-rfi/src/components/appVariants/index.js
@@ -83,7 +83,7 @@ const variants = {
         To learn more about ASU or a specific program, fill out the form below
         then check your email.
       </p>
-      <Campus gaData={gaData1Of2} />
+      <Campus autoFocus gaData={gaData1Of2} />
       <CareerAndStudentType gaData={gaData1Of2} />
       <Interest1 gaData={gaData1Of2} />
       <Interest2 gaData={gaData1Of2} />
--- a/packages/app-rfi/src/components/steps/questions/Campus.js
+++ b/packages/app-rfi/src/components/steps/questions/Campus.js
@@ -8,7 +8,8 @@ import { RfiSelect } from "../../controls";
 /**
  * @param {{ gaData: import("../../../../../../shared/services/googleAnalytics").GAEventObject}} props
  */
-export const Campus = ({ gaData }) => {
+// eslint-disable-next-line react/prop-types
+export const Campus = ({ gaData, autoFocus}) => {
   const label = "Which applies to you?";
   const name = "Campus";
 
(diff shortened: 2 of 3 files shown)


packages/app-rfi/src/components/appVariants/index.js
```diff
@@ -83,7 +83,7 @@ const variants = {
         To learn more about ASU or a specific program, fill out the form below
         then check your email.
       </p>
-      <Campus gaData={gaData1Of2} />
```
scott-williams-az: This would cause autofocus on page load. I will post a code sample of what I'm thinking will be a better solution.

scott-williams-az (changes requested):

packages/app-rfi/src/components/appVariants/index.js
scott-williams-az: These forms are usually at the bottom of a long page so we need to make sure we do not focus on page load. 

You will need to test this and figure out if I missed anything.

We can also probably also get rid of the autofocus prop from other areas of the code.

What do you think of this? (I also added the errors as optional, but I think it is worth having. Just comment out that line if we aren't sure we need it.)

### Add

```javascript
import { useFormikContext } from "formik";
``` 

<details><summary>... (Open to see context)</summary>

```javascript
import PropTypes from "prop-types";
import React, { useEffect, useRef } from "react";
``` 

</details> 




```javascript
const getForkicFieldElements = (elem, obj) => {
  const arr = Object.keys(obj);
  if (arr.length === 0) {
    return [];
  }

  // querySelectorAll will ensure the first index is the first element on the current rendering of the page
  return elem.querySelectorAll(arr.map(n => `[name="${n}"]`).join(","));
};

``` 

<details><summary>...</summary>




```javascript

const RfiMainForm = ({ rfiImage, children }) => {

``` 

</details> 




```javascript

  const { submitCount, values, errors } = useFormikContext();
  const element = useRef(null);

  useEffect(
    () => {
      if (
        // prevent focusing on first render
        submitCount > 0 &&
        // @ts-ignore
        typeof element?.current?.querySelectorAll === "function"
      ) {
        const nodes = [
          // if there are errors, focus on the first one
          ...getForkicFieldElements(element.current, errors),
          // if there are no errors, focus on the first field
          ...getForkicFieldElements(element.current, values),
        ];
        nodes[0].focus();
      }
    },
    // Must also watch children prop since the Previous button click does not change submitCount
    [submitCount, children]
  );


``` 
<details><summary>...</summary>




```javascript

  return (
    <div className="container rfi-container-inner">
      <div className="row">
        <div className="col col-12 ">
          <div className="uds-image-text-block-container">
            <div className="uds-image-text-block-image-container">
              <img src={rfiImage} className="img-fluid" alt="" />
            </div>

``` 

</details> 




```javascript

            <div ref={element} className="uds-image-text-block-text-container">

``` 
<details><summary>...</summary>




```javascript

              {children}
            </div>
          </div>
        </div>
      </div>
    </div>
  );
};

RfiMainForm.propTypes = {
  rfiImage: PropTypes.string.isRequired,
  children: PropTypes.node.isRequired,
};

export { RfiMainForm };
</details> 
```

scott-williams-az: The changes I mentioned were for the `RfiMainForm.js` (I failed to mention before)

Also I found a possible solution to the bug we saw during standup with the focus getting stuck.


 
`const { step } = useRfiContext();`
...

~`// Must also watch step prop since the Previous button click does not change submitCount`~
`// Must also watch children prop since the Previous button click does not change submitCount`
~`[submitCount, children]`~
`[submitCount, step]`

If we do not implement auto focus to error, then we can just watch `[step]` instead of `[submitCount, step]`

juanmitriatti:
Hello @scott-williams-az 
Good morning
I just added some code in order to make it work. 
Please test it and let me know whenever you’ve a chance .
Thanks you for your feedback.

scott-williams-az (approved):

davidornelas11 (approved):
Tested locally and focus follows expected behavior. Tested in webspark as well

PR #1515, opened 2025-04-10
Uds 1958

scott-williams-az:
- [JIRA ticket](https://asudev.jira.com/browse/UDS-1958)


packages/unity-bootstrap-theme/vite.config.js
```diff
@@ -50,26 +50,30 @@ const c = {
       entry: [
         resolve(__dirname, "src/scss/unity-bootstrap-theme.bundle.scss"),
         resolve(__dirname, "src/scss/unity-bootstrap-theme.scss"),
+        resolve(__dirname, "src/scss/unity-bootstrap-header-footer.scss"),
         resolve(__dirname, "src/scss/unity-bootstrap-header.scss"),
         resolve(__dirname, "src/scss/unity-bootstrap-footer.scss"),
         resolve(__dirname, "src/js/global-header.js"),
         resolve(__dirname, "src/js/data-layer.js"),
-        resolve(__dirname, "../../node_modules/bootstrap/js/index.esm.js"),
+        // resolve(__dirname, "../../node_modules/bootstrap/js/index.esm.js"),
```
davidornelas11: This can just be removed since we just copy the bootstrap umd over now.

packages/unity-bootstrap-theme/vite.config.js
```diff
@@ -50,26 +50,30 @@ const c = {
...
       treeshake: true,
       output: {
+        globals: {
+          "chart.js": "Chart",
+        },
         entryFileNames: chunkInfo => {
           if (chunkInfo.name.includes("index.esm")) {
-            return "js/bootstrap.bundle.min.[format]";
+            return "js/[format]/bootstrap.bundle.min.js";
           }
```
davidornelas11: This can also be removed now since the copy-bootstrap-umd-to-dist plugin up top already moves the bootstrap umd over to the dist file. This just overwrites that anyways

packages/unity-bootstrap-theme/package.json
```diff
@@ -77,14 +79,6 @@
...
   "exports": {
-    "./js/global-header.js": {
-      "module": "./src/js/storybook-global-header.js",
-      "default": "./src/js/global-header.js"
-    },
-    "./js/data-layer.js": {
-      "module": "./src/js/storybook-data-layer.js",
-      "default": "./src/js/data-layer.js"
-    },
     "./*": "./*",
```
davidornelas11, line 82: We just have to make sure we update our README to provide any info needed for imprting certain files

packages/unity-react-core/src/components/Card/Card.jsx
```diff
@@ -154,44 +154,26 @@ Card.propTypes = {
   cardLink: PropTypes.string,
 };
```
davidornelas11, line 156: Nice job removing the default props on these. I think there is a ticket for this so we might be able to close that too. I will check

davidornelas11 (dismissed):
Looks good. Just a few things I noticed

packages/unity-bootstrap-theme/vite.config.bundle.js
```diff
@@ -1,85 +1,53 @@
-import react from "@vitejs/plugin-react";
-import path, { resolve } from "path";
 import fs from "fs";
-import { defineConfig, transformWithEsbuild } from "vite";
+import { resolve } from "path";
+import { defineConfig } from "vite";
 
-import pkg from "./package.json";
-/** @typedef {import('vite').UserConfig} UserConfig */
+import baseConfig from "./vite.config.js";
```
davidornelas11, line 5: Good idea reusing the build config

packages/unity-bootstrap-theme/vite.config.bundle.js
```diff
@@ -1,85 +1,53 @@
...
-      treeshake: true,
-      output: {
-        globals: {
-          "chart.js": "Chart",
+    return {
+      ...baseConfig,
+      plugins: [
+        {
+          name: "copy",
+          writeBundle() {
```
davidornelas11: One small caveat here is that this runs 3 times, one for each format. The closeBundle hook only runs once after the entire build process runs. This is not an issue in local but in the CI that might be underpowered and doing this dozens of times, it might add up. I console logged the time it takes to process and its a difference of 500-800ms for closeBundle vs 2-4 seconds for writeBundle. Might be different on Jenkins. Screenshot is only how many times it runs
Screenshot 2025-04-24 at 2 43 47 PM
Screenshot 2025-04-24 at 2 44 14 PM

davidornelas11: If we were doing anything with the build itself then we would have to call bundle.close() but we should be fine not needing to call it

davidornelas11 (approved):

PR #1534, opened 2025-05-12
chore: add precommit hook and update linted files

davidornelas11:
add precommit hook and update linted files

## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2004)

### DIFF
--- a/.husky/prepare-commit-msg
+++ b/.husky/prepare-commit-msg
@@ -8,4 +8,16 @@ if [ -n "$2" ]; then
   exit 0
 fi
 
+echo "Running linter..."
+yarn lint
+
+# Check linter exit status
+if [ $? -ne 0 ]; then
+  echo "Linting failed. Please fix lint errors before committing."
+  exit 1
+else
+  echo "Adding linted files to staging..."
+  git add .
+fi
+
 exec < /dev/tty && git cz --hook || true
--- a/.husky/prepare-commit-msg
+++ b/.husky/prepare-commit-msg
@@ -8,16 +8,4 @@ if [ -n "$2" ]; then
   exit 0
 fi
 
-echo "Running linter..."
-yarn lint
-
-# Check linter exit status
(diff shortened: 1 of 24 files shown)


.husky/prepare-commit-msg
```diff
@@ -8,4 +8,16 @@ if [ -n "$2" ]; then
...
+echo "Running linter..."
+yarn lint
+
+# Check linter exit status
+if [ $? -ne 0 ]; then
+  echo "Linting failed. Please fix lint errors before committing."
+  exit 1
+else
+  echo "Adding linted files to staging..."
+  git add .
```
scott-williams-az: We do not want to add all files, in an automatic way without the contributor knowing. This should be the developers choice to include changes. 

1. It is possible for a contributor to have changes they do not want to include with the commit.
2. I think this prevents contributor from making smaller commits to group similar changes.

scott-williams-az (commented):
I was thinking this hook would be a pre-push hook. (can we include tests too? it will save us from pushing a failed build to jenkins)

Sometimes while developing locally I just want to commit my changes. I may have code I know is not ready for a PR but I need to make a checkpoint.


We should as much as possible make sure all paths are covered.
`git commit` calls the hook
`yarn commit` or calling `cz` does not call prepare-commit-msg

.husky/prepare-commit-msg
davidornelas11: So should this be a prepush hook instead? If it's a prepush hook, then commitizen shouldn't be called at all because that's already been done before they ran git push, right?

scott-williams-az: Correct

davidornelas11: OK made that change

scott-williams-az (dismissed):

scott-williams-az (approved):

PR #1564, opened 2025-07-16
UDS-1970: feat(unity-bootstrap-theme): added mixin to expand clickable area

juanmitriatti:
added mixin to expand clickable area for btn

UDS-1970

## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1970)

### DIFF
--- a/packages/unity-bootstrap-theme/src/scss/extends/_buttons.scss
+++ b/packages/unity-bootstrap-theme/src/scss/extends/_buttons.scss
@@ -5,15 +5,12 @@
     1rem;
 }
 
-@mixin clickable-area($min-width: 80px, $min-height: 44px) {
+@mixin clickable-area() {
   position: relative;
   &::before {
     content: "";
     position: absolute;
-    min-width: $min-width;
-    min-height: $min-height;
-    z-index: -1;
-    inset: 0;
+    inset: -10px;
     margin: auto;
   }
 }
@@ -24,6 +21,8 @@
   font-weight: bold;
   white-space: nowrap;
   width: max-content !important;
+  @include clickable-area();
+
   &::first-letter {
     text-transform: uppercase;
   }
@@ -33,14 +32,11 @@
   }
   &.btn-md {
     @include btn-md;
-    @include clickable-area(80px, 44px);
   }
   &.btn-sm {
     font-size: 0.75rem;
     padding: $uds-component-button-padding-y-small
       $uds-component-button-padding-x-small;
-
-    @include clickable-area(64px, 44px);
   }
   &.btn-tag {
     font-size: 0.75rem;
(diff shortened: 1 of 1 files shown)


packages/unity-bootstrap-theme/src/scss/extends/_buttons.scss
```diff
@@ -5,6 +5,19 @@
...
 }
 
+@mixin clickable-area($min-width: 80px, $min-height: 44px) {
+  position: relative;
+  &::before {
+    content: "";
+    position: absolute;
+    min-width: $min-width;
+    min-height: $min-height;
+    z-index: -1;
```
scott-williams-az: Since we are using `::before`, the text will naturally have a higher z-index, since it comes after `::before` and we should not use z-index:-1 since this will push the `::before` content behind an element with position: relative and will prevent the clickable-area from working

packages/unity-bootstrap-theme/src/scss/extends/_buttons.scss
```diff
@@ -5,6 +5,19 @@
...
+    inset: 0;
+    margin: auto;
+  }
+}
```
scott-williams-az, line 16: Image

```
  &::before {
    content: "";
    position: absolute;
    min-width: $min-width; // is not needed, but doesn't hurt
    min-height: $min-height;
    // z-index: -1;
    inset: 0;
    margin: auto;
    right: min(0px, calc(50% - #{$min-width} / 2));
    left: min(0px, calc(50% - #{$min-width} / 2));


    background-color: #f002; // visual testing only
  }
```
The code above equally expands the clickable area in all directions, The only issue is the overlap with the 1 character buttons, but I think this is an extreme scenario

packages/unity-bootstrap-theme/src/scss/extends/_buttons.scss
```diff
@@ -5,6 +5,19 @@
...
+
 // Extra styles that don't have bootstrap variables to override
 .btn {
```
scott-williams-az, line 19: I the solution needs to be on the .btn class since buttons are responsive and regular buttons inherit a small class with a narrow viewport.

Image

packages/unity-bootstrap-theme/src/scss/extends/_buttons.scss
```diff
@@ -5,6 +5,19 @@
     1rem;
 }
 
+@mixin clickable-area($min-width: 80px, $min-height: 44px) {
+  position: relative;
+  &::before {
+    content: "";
+    position: absolute;
+    min-width: $min-width;
```
scott-williams-az: Setting min-width only increases the clickable area in one direction 

Image

scott-williams-az (dismissed):

juanmitriatti:
Hi @scott-williams-az  @davidornelas11 
This is ready for review again. 
Thanks in advance.

davidornelas11 (approved):
Looks good and expands the click area

PR #1577, opened 2025-09-08
Uds 2036 - update multiple card render for unity-react-core cards

davidornelas11:
This adds the options for rendering multiple cards for the ranking cards and Card component in unity-react-core. Webspark will have to be updated to use this but the default/old behavior is still there and users will have to intentinally use the new prop to use the new render feature
## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2036)

### DIFF
--- a/packages/unity-react-core/src/components/Card/Card.jsx
+++ b/packages/unity-react-core/src/components/Card/Card.jsx
@@ -65,30 +65,40 @@ export const Card = ({
     };
 
     return (
-        <div className={classNames("uds-card-arrangement-card-container", "auto-arrangement", getColumnClass())}>
-          {cards.map((card, index) => (
-            <BaseCard
-              key={index}
-              type={card.type || type}
-              width={card.width || width}
-              horizontal={card.horizontal !== undefined ? card.horizontal : horizontal}
-              image={card.image}
-              imageAltText={card.imageAltText}
-              title={card.title}
-              icon={card.icon}
-              body={card.body}
-              eventFormat={card.eventFormat || eventFormat}
-              eventLocation={card.eventLocation}
-              eventTime={card.eventTime}
-              buttons={card.buttons}
-              linkLabel={card.linkLabel}
-              linkUrl={card.linkUrl}
-              tags={card.tags}
-              showBorders={card.showBorders !== undefined ? card.showBorders : showBorders}
-              cardLink={card.cardLink}
-            />
-          ))}
-        </div>
+      <div
+        className={classNames(
+          "uds-card-arrangement-card-container",
+          "auto-arrangement",
+          getColumnClass()
+        )}
+      >
+        {cards.map((card, index) => (
+          <BaseCard
+            key={index}
+            type={card.type || type}
+            width={card.width || width}
+            horizontal={
+              card.horizontal !== undefined ? card.horizontal : horizontal
+            }
+            image={card.image}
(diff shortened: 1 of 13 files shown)


mlsamuelson (changes requested):
@davidornelas11 I like this approach, it moves us forward without introducing a breaking change. Nice. Appears it will be adaptive for use with multiple column types based on where it's placed.

Only concern at this point is problematic rendering when images have different heights: https://unity-uds-staging.s3.us-west-2.amazonaws.com/pr-1577/@asu/unity-react-core/index.html?path=/story/components-image--three-images-arrangement 

Screenshot 2025-09-12 at 2 24 53 PM

davidornelas11:
Per brand team, a 4:3 aspect ratio will be enforced with the styles when the card-arrangement is used. No changes to previous behavior. Webspark already crops and takes care of this so will only affect nonwebspark users

packages/unity-react-core/src/components/Card/Card.jsx
```diff
@@ -152,6 +214,57 @@ Card.propTypes = {
    * Card link
    */
   cardLink: PropTypes.string,
+  /**
+   * Array of card objects for rendering multiple cards
+   */
+  cards: PropTypes.arrayOf(
```
scott-williams-az, line 220: Cards has same structure as BaseCard, might need to rearrange the file, but can we reuse the prop structure

cards: PropTypes.arrayOf(PropTypes.shape(BaseCard.propTypes))

packages/unity-react-core/src/components/Card/Card.jsx
```diff
@@ -46,7 +46,69 @@ export const Card = ({
   tags,
   showBorders = true,
   cardLink,
+  cards = [],
+  columns = "0",
```
scott-williams-az, line 50: I think we are moving towards responsive design, and we should not allow columns to be a property. Columns should be responsive and allow the card to adjust between the min and max widths provided by the brand team.

packages/unity-react-core/src/components/Card/Card.jsx
```diff
@@ -46,7 +46,69 @@ export const Card = ({
   tags,
   showBorders = true,
   cardLink,
+  cards = [],
```
scott-williams-az, line 49: Card feels like it is doing too much. Should we have a card-arrangement component that takes an array of 1 or more cards?

scott-williams-az: I am ok with this temporarily if we have to keep it for WS2 purposes, but think it will cause issue down the line since it is a strange pattern

packages/unity-react-core/src/components/Card/Card.jsx
```diff
@@ -46,7 +46,69 @@ export const Card = ({
...
+  columns = "0",
+  layout = "auto"
```
scott-williams-az, line 51: Does `auto` mean horizontal?
Instead of `layout` default to `auto` can we change `layout` prop to `vertical` = `true` or `forceVertical` = `true`?

scott-williams-az: I see we already have  horizontal prop. these props are getting confusing

davidornelas11:
Closing in favor of https://github.com/ASU/asu-unity-stack/pull/1603

PR #1596, opened 2025-10-23
fix(app-degree-pages): refactor defaultProps to use JS default params

oatkar:
### Description

 Refactoring defaultProps to use JS default params

## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1735?atlOrigin=eyJpIjoiNjMxMTI0NmJiZDhjNDg5MjljMDkxNmNiNWJjNDRlMDQiLCJwIjoiaiJ9)


packages/app-rfi/src/components/controls/controls-helpers.js
```diff
@@ -15,7 +15,7 @@ const GaEventPropTypes = {
   text: PropTypes.string,
 };
 
-const RfiLabel = ({ label, name, id, requiredIcon }) => (
+const RfiLabel = ({ label, name, id = undefined, requiredIcon = undefined }) => (
```
scott-williams-az: As mentioned in another review comment, we do not need to add ` = undefined`

For clarity IF we wanted a default prop for requiredIcon we would use `requiredIcon = false` since it is a boolean.

But this is a small component, there is no confusion logic and we can see and understand the code quickly.

I am ok either way, if you want to leave `requiredIcon = false`

packages/app-rfi/src/components/controls/controls-helpers.js
```diff
@@ -15,7 +15,7 @@ const GaEventPropTypes = {
...
   <label htmlFor={id || name}>
     {requiredIcon && (
```
scott-williams-az: I just wanted to point out a common pattern you might see in other places

`{requiredIcon && (` could also look like `{!!requiredIcon && (`

In a scenario where variable `requiredIcon` is `undefined`.
`requiredIcon` computes to `undefined` which is `falsey`
`!requiredIcon` the first `!` changes the `falsey` to `true` (value is inverted but it is now a boolean type)
`!!requiredIcon` the second `!` inverts the value again, except it is no longer falsey (value is false and type is a boolean)

packages/component-footer/src/components/Social/index.js
```diff
@@ -28,7 +28,7 @@ const DEFAULT_GA_EVENT = {
  * @returns {JSX.Element}
  */
 
-const Social = ({ social: { logoUrl, unitLogo, mediaLinks } }) => {
+const Social = ({ social: { logoUrl, unitLogo = endorsedLogo, mediaLinks = {facebook: "", twitter: "", linkedIn: "", instagram: "", youtube: ""} } }) => {
```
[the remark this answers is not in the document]
davidornelas11: @oatkar This is actually a decent suggestion from copilot. Can you create a separate const with those values in medialinks and just assign it to mediaLinks

davidornelas11 (changes requested):
Once change requested by copilot

mlsamuelson (commented):
Would love to see the copilot suggestions evaluated, and then we can re-review.

davidornelas11:
@oatkar Pinging you for some of the changes requested

davidornelas11:
@oatkar Just some merge conflicts and this should be ready

davidornelas11 (approved):

PR #1612, opened 2025-12-22
feat(unity-bootstrap-theme): add new button styles

davidornelas11:
Add new button styles and deprecate old ones
## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1610)


mlsamuelson (changes requested):
Requesting changes. 

As we discussed, while these changes look good, if we make the "filled" variant the default behavior, then we can skip using a separate "UDSButton" component and instead "enhance in place" the Button component, adding the new props and assuming the default (classic button) when they're not present.

This will also allow for us to avoid needing to eventually update every component that relies on the Button component in UDS, and in WS2 CKEditor we can simply add the new button variations **without** needing to rewrite parts of the Button plugin and include mapping logic for existing content using buttons.

mlsamuelson:
Oh, also see if the double counts on the datalayer can be addressed as well.

mlsamuelson (approved):

PR #1621, opened 2026-01-28
fix(unity-react-core): fix image card rendering in card arrangement

davidornelas11:
fix image card rendering in card arrangement
See screenshots below of what it looks like in webspark
Screenshot 2026-01-28 at 12 02 20 PM
Screenshot 2026-01-28 at 12 03 59 PM
Screenshot 2026-01-28 at 12 04 13 PM
Screenshot 2026-01-28 at 12 04 22 PM

### DIFF
--- a/packages/unity-react-core/src/components/CardArrangement/CardArrangement.jsx
+++ b/packages/unity-react-core/src/components/CardArrangement/CardArrangement.jsx
@@ -74,8 +74,9 @@ export const CardArrangement = ({ cards, cardType = "card", columns }) => {
 
 CardArrangement.propTypes = {
   /**
-   * Array of card objects to render. Each object should contain props for the Card or RankingCard component.
-   * Cards will wrap naturally based on their intrinsic min/max widths.
+   * Array of card objects to render. Each object should contain props for the
+   * Card or RankingCard component. Cards will wrap naturally based on their
+   * intrinsic min/max widths.
    */
   cards: PropTypes.arrayOf(PropTypes.object).isRequired,
   /**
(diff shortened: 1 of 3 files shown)


mlsamuelson (approved):

mlsamuelson (changes requested):
I  don’t want to hold this up, a few things looking in Storybook:
I’m guessing that using the controls for `cardType` doesn’t work real well because the different types require different `card` props. What do you think about providing an image cards story so that can be verified in Storybook?

We have two different card type TYPEs don’t we?
The `cardType` prop at the top level that can be one of `ranking`, `card` or `image`.
and then the `cards.[].type` that can be one of `default`, `degree`, `event`, `story`.

Okay, so we have
* ranking cards
* image cards
* cards which can be
  * default
  * degree
  * event
  * story

That look right to you, [@dornela3]

Let's spell that out in the docs page so we add some clarity to help make it clear. https://unity-uds-staging.s3.us-west-2.amazonaws.com/pr-1621/@asu/unity-react-core/index.html?path=/docs/components-card-arrangement--docs

The `cards` prop description could become:
`Array of card objects to render. Each object should contain props for the Card, Image or RankingCard component. Cards will wrap naturally based on their intrinsic min/max widths. When using the Card component, cards can supply a type of "default", "degree", "event" or "story".`

Or something like that.


And then add at least 1 Image cards story so we have an example.

mlsamuelson (dismissed):
Looks like build issues need to be resolved.

mlsamuelson (approved):

mlsamuelson:
Ticket for this effort: https://asudev.jira.com/browse/WS2-2868

PR #1655, opened 2026-03-17
fix(unity-bootstrap-theme): fix visited links issue in header

davidornelas11:
fix visited links issue in header

### DIFF
--- a/packages/unity-bootstrap-theme/src/scss/extends/_misc.scss
+++ b/packages/unity-bootstrap-theme/src/scss/extends/_misc.scss
@@ -10,7 +10,15 @@
 }
 
 // Global link visited state
-a:visited:not(.btn):not(.nav-link):not(.page-link):not(.card-link) {
+a:visited:not(
+    .btn,
+    .nav-link,
+    .page-link,
+    .card-link,
+    header *,
+    footer *,
+    .uds-grid-links *
+  ) {
   color: $uds-color-base-darkmaroon;
 }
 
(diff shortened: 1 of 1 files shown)


mlsamuelson (dismissed):

mlsamuelson (approved):
Verified in the static test site. Refactor o the rules and updates looks good. Not able to identify any other cases where we need to adjust, but this can grow if needed.

mlsamuelson:
Shared for additional reviews, but this is timely for a numbe of uses. Can adjust with additional updates if needed.

packages/unity-bootstrap-theme/src/scss/extends/_misc.scss
```diff
@@ -10,7 +10,15 @@
...
 
 // Global link visited state
-a:visited:not(.btn):not(.nav-link):not(.page-link):not(.card-link) {
+a:visited:not(
+    .btn,
+    .nav-link,
+    .page-link,
+    .card-link,
+    header *,
+    footer *,
```
SteveRyan-ASU, line 19: Wondering if targeting the general `<header>` and `<footer>` elements here will wind up being too generic and cause some additional style bleed out in places? Its possible in some constructions to reuse these elements on the page.

Maybe consider scoping it specifically to the markers in component-header-footer?
``` 
a:visited:not(
   #asuHeader *,
   #wrapper-endorsed-footer *,
   #wrapper-footer-columns *,
   #wrapper-footer-innovation *,
   #wrapper-footer-colophon *,
   ...
) {
   color: $uds-color-base-darkmaroon;
}

SteveRyan-ASU (commented):
Here's maybe an alternate idea?

PR #1660, opened 2026-03-20
chore: add ai agent instructions for unity-react-core

davidornelas11:
add ai agent instructions for unity-react-core

### DIFF
--- a/.github/instructions/unity-react-core.instructions.md
+++ b/.github/instructions/unity-react-core.instructions.md
@@ -1,0 +1,90 @@
+---
+applyTo: "**/unity-react-core/**"
+---
+
+# Unity React Core — Copilot Agent Instructions
+
+## Package Overview
+
+All components in this package are part of the `unity-react-core` monorepo package. They are paired with a **Storybook HTML addon** that allows consumers to copy raw HTML markup and use it independently — without React — while achieving identical visual output. This constraint shapes every architectural and styling decision below.
+
+---
+
+## Styling: Always Use `unity-bootstrap-theme` Variables
+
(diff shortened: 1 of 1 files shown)


.github/instructions/unity-react-core.instructions.md
```diff
@@ -0,0 +1,90 @@
...
+  // TODO: No variable found for this value in _custom-asu-variables.scss — confirm with design.
+  color: $your-value-here;
+  ```
+- Variable usage applies to component SCSS files, inline style props, and any CSS-in-JS patterns. Prefer SCSS.
+
+---
+
+## HTML Parity Requirement (Storybook HTML Addon)
+
+Every component must render its **full, correct HTML structure on the first render** — very little exceptions. The Storybook HTML addon captures the static markup so consumers can copy it and use it with only the compiled CSS, completely outside of React.
```
juanmitriatti, line 33: Hey @davidornelas11  
Just asking, is not it better to use "very few exceptions" or better yet, "with very few exceptions"?
Instead of "very little exceptions".

juanmitriatti:
@davidornelas11  Should not we mention accesibility?  I think it would be great if  the checklist  includes something like:
  - [ ] Interactive elements have appropriate ARIA attributes (`aria-label`, `aria-expanded`,
  `role`, etc.)
  - [ ] Component is keyboard navigable
  Since the existing codebase already follows this (e.g., Button has ariaLabel prop), so the instructions
  should reinforce it.

.github/instructions/unity-react-core.instructions.md
mlsamuelson: That would be better wording. I can be adjusted later if needed. Not considering it blocking.

mlsamuelson (approved):
Approving. Can be adjusted as we move forward if needed.

mlsamuelson:
@juanmitriatti I'm considering this a good baseline to start from. Feel free to submit additional PRs to propose expanding on the agent instructions.

PR #1661, opened 2026-03-25
chore: upgrade to storybook10, vite6, vitest and msw

davidornelas11:
This PR updates all packages to Storybook 10.3+ except component-header or component-footer. Those are now deleted


scott-williams-az (approved):

mlsamuelson (commented):
This looks good to me.

Minor issue with error in pagination story in Bootstrap mode:
`Rendered more hooks than during the previous render.`
But that looks like it's a SB only rendering issue that wouldn't affect users of the packages.

PR #1667, opened 2026-04-01
fix(unity-bootstrap-theme): fix anchor menu js

davidornelas11:
Anchor menu fix to fix the focus order as well as adds correct hash in url bar. Also adds check for unity-react-core react anchor menu since that handles its own logic, currently used in app degree detail page. Eventually we want to remove that logic in the unity-react-core anchor menu but that will be a breaking change. 

I tried using scroll padding and scroll margin only but when I tested it on Fulton engineering's wordpress site, it did not work as expected. This was tested on multiple sites and seems to be the best solution I have used so far
## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1717)

### DIFF
--- a/packages/unity-bootstrap-theme/src/js/anchor-menu.js
+++ b/packages/unity-bootstrap-theme/src/js/anchor-menu.js
@@ -15,9 +15,9 @@ function initAnchorMenu() {
   const globalHeader = document.getElementById(globalHeaderId);
   const navbar = document.getElementById("uds-anchor-menu");
   if (!navbar || !globalHeader) {
-    console.warn(
-      "Anchor menu initialization failed: required elements not found"
-    );
+    return;
+  }
+  if (Array.from(navbar.classList).some(cls => cls.startsWith("sc-"))) {
     return;
   }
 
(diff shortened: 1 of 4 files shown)


packages/unity-bootstrap-theme/src/js/anchor-menu.js
```diff
@@ -203,10 +203,26 @@ function initAnchorMenu() {
...
+      // For the first anchor item, skip scrolling if the section top is
+      // already clearly visible in the viewport (above the midpoint).
+      const isFirstAnchor = anchor === anchors[0];
+      if (isFirstAnchor) {
+        const headerBottom = globalHeader.getBoundingClientRect().bottom;
+        const navbarHeight = navbar.offsetHeight;
+        const topOffset = headerBottom + navbarHeight;
+        const targetTop = anchorTarget.getBoundingClientRect().top;
+        const viewportMid = window.innerHeight / 2;
+        console.log("viewportmid", viewportMid, "targetTop", targetTop);
```
mlsamuelson: Can we remove this console.log?

mlsamuelson (dismissed):
Are there limitations on how should I be able to verify this one in storybook? I haven't dug into the stories and how their assets are loaded in storybook since they're not in the PR, but what I see is:

* asu-react-core react anchor works
* asu-react-core bootstrap anchor doesn't
* unity-bootstrap-theme anchor works

Given that the pr is for unity-bootstrap-theme, I thinking this is a pass, but wondering if there's any way the asu-react-core fix can happen, or is does that involve the future breaking change?

scott-williams-az (commented):
[react](https://unity-uds-staging.s3.us-west-2.amazonaws.com/pr-1667/@asu/unity-react-core/iframe.html?id=components-anchormenu--default&viewMode=story&globals=framework%3Areact)
[bootstrap](https://unity-uds-staging.s3.us-west-2.amazonaws.com/pr-1667/@asu/unity-react-core/iframe.html?id=components-anchormenu--default&viewMode=story&globals=framework%3Abootstrap)

Bootstrap version is good.

Possible a11y issues with react version

- when clicking (or keyboard press enter) the anchor does become visible, but the focus does not move to the anchor point (I believe it should)
- the url is not updated with the `url#anchor-name`

davidornelas11:
> [react](https://unity-uds-staging.s3.us-west-2.amazonaws.com/pr-1667/@asu/unity-react-core/iframe.html?id=components-anchormenu--default&viewMode=story&globals=framework%3Areact) [bootstrap](https://unity-uds-staging.s3.us-west-2.amazonaws.com/pr-1667/@asu/unity-react-core/iframe.html?id=components-anchormenu--default&viewMode=story&globals=framework%3Abootstrap)
> 
> Bootstrap version is good.
> 
> Possible a11y issues with react version
> 
> * when clicking (or keyboard press enter) the anchor does become visible, but the focus does not move to the anchor point (I believe it should)
> * the url is not updated with the `url#anchor-name`

I'm thinking we can merge this one then update the react anchor menu next. The near goal is to just remove all logic from the anchor menu in react and only rely on the anchor-menu.js in unity-bootstrap-theme. What do you think?

scott-williams-az (approved):

PR #1670, opened 2026-04-02
feat: add pre-commit check for merge conflict markers

scott-williams-az:
### Description
After resolving merge conflicts with a large number of resolutions, We need to make sure we do not miss any resolutions. This has happened a few times, and we want to make sure we do not accidentally include a broken file

Write pre-commit hook to make sure staged files do not include 

```
<<<<<<< HEAD
...
=======
...
>>>>>>> [Commit hash]
```
### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2142)

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,31 @@
 #!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+
+# Check for merge conflict markers
+# This hook prevents commits that contain unresolved merge conflicts.
+# It checks the staged changes for common conflict markers
+# (e.g., <<<<<<<, =======, >>>>>>>) and blocks the commit if any are found.
+
+# This file will always fail since it contains the conflict markers pattern,
+# We will still check this file for conflict markers to prevent committing
+# unresolved conflicts.
+
+# In order to intentionallyupdate this file, create a commit with only this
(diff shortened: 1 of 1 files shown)


.husky/pre-commit
```diff
@@ -1 +1,31 @@
...
+if [ "$SKIP_CONFLICT_CHECK" != "1" ]; then
+    conflicts=$(git diff --cached --name-only | xargs grep -H -n -E "<{7} HEAD|={7}|>{7} " 2>/dev/null)
+    if [ ! -z "$conflicts" ]; then
+        echo "Error: Merge conflict markers found in staged files."
+        echo "\nConflicts found:"
+        echo "$conflicts" | while IFS= read -r line; do
+            echo "  → $(echo "$line" | sed 's/\([0-9]\):\([<>=]\)/\1 \2/')"
+        done
+        echo "\nPlease resolve all merge conflicts before committing."
+        exit 1
```
scott-williams-az, line 26: There is a minor issue with condition, need to make small adjustment

scott-williams-az: The issue was an accident line paste on line 2, I removed line and replaced with line to skip hook for CI pipeline.

davidornelas11 (approved):
looks good and works when I tested it locally

PR #1674, opened 2026-04-06
fix(unity-bootstrap-theme): blockquote css update for maxwidth standards

spaceperson:
## Checklist

### Links

### DIFF
--- a/packages/unity-bootstrap-theme/src/scss/extends/_blockquotes.scss
+++ b/packages/unity-bootstrap-theme/src/scss/extends/_blockquotes.scss
@@ -11,9 +11,9 @@ blockquote {
   border-left: none;
   font-weight: inherit;
   padding-left: 0;
-  max-width: none;
+  max-width: 772px;
   p {
-    max-width: unset;
+    max-width: 700px;
     overflow-wrap: break-word;
     word-wrap: break-word;
     word-break: break-word;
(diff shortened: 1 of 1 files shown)


davidornelas11 (changes requested):
the `div` with the class of `uds-blockquote` should have a left padding of 0. The blockquote itself has padding left set to 0 already but that also has to be applied to the parent div with the aforementioned class as well

packages/unity-bootstrap-theme/src/scss/extends/_blockquotes.scss
```diff
@@ -45,6 +45,7 @@ blockquote:before {
   font-weight: 700;
 
   padding: $uds-size-spacing-4;
+  padding-left: 0;
```
scott-williams-az, line 49: The JIRA ticket says to remove padding from left and right. 

Also make sure you test these stories

https://unity-uds-staging.s3.us-west-2.amazonaws.com/pr-1674/@asu/unity-bootstrap-theme/index.html?path=/story/atoms-blockquotes-and-testimonials-examples--testimonials-no-image.

scott-williams-az (changes requested):

packages/unity-bootstrap-theme/src/scss/extends/_blockquotes.scss
spaceperson: I've included padding-right: 0; into the source code.

bebaps (approved):
Ran locally and looks good. I also manually applied these changes on a WS site, and when this gets moved over, there is a small amount of CSS that will need to be removed since this PR now includes the styles.

davidornelas11 (approved):

PR #1676, opened 2026-04-07
Update news cards

davidornelas11:
## Checklist


packages/unity-react-core/src/components/ComponentCarousel/components/CardCarousel/CardCarousel.jsx
```diff
@@ -68,6 +68,7 @@ const htmlTemplate = (
       linkLabel={linkLabel}
       linkUrl={linkUrl || buttons?.[0]?.href}
       tags={tags}
+      cardLink={linkUrl || buttons?.[0]?.href}
```
scott-williams-az, line 71: This probably happened over time where we added one feature (prop) at a time, but this component is asking for things it already has. I do not have any problems with the code change, but these extra props made me take more time because I wanted to make sure it wasn't a mistake.

scott-williams-az (approved):

PR #1680, opened 2026-04-10
fix(unity-bootstrap-theme): prevent tab vertical scroll

scott-williams-az:
### Description

Tabbed content block malfunctioning

### Links

- [JIRA ticket](https://asudev.jira.com/browse/WS2-2939)

### DIFF
--- a/packages/unity-bootstrap-theme/src/js/tabbed-panels.js
+++ b/packages/unity-bootstrap-theme/src/js/tabbed-panels.js
@@ -25,18 +25,19 @@ function initTabbedPanels() {
       e.target.focus();
     }
   };
-  const scrollIntoViewHorizontally = (container, child) => {
-    const childOffsetRight = child.offsetLeft + child.offsetWidth;
+  const scrollIntoViewHorizontally = tab => {
+    const container = tab.parentElement;
+    const tabOffsetRight = tab.offsetLeft + tab.offsetWidth;
     const containerScrollRight = container.scrollLeft + container.offsetWidth;
 
-    // is child behind (left)
-    if (container.scrollLeft > child.offsetLeft) {
-      container.scrollLeft = child.offsetLeft;
+    // is tab behind (left)
+    if (container.scrollLeft > tab.offsetLeft) {
+      container.scrollLeft = tab.offsetLeft;
     }
 
-    // is child ahead (right)
-    if (containerScrollRight < childOffsetRight) {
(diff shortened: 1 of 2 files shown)


packages/unity-bootstrap-theme/src/js/tabbed-panels.js
```diff
@@ -25,6 +25,21 @@ function initTabbedPanels() {
...
     }
   };
+  const scrollIntoViewHorizontally = tab => {
+    const container = tab.parentElement;
+    const tabOffsetRight = tab.offsetLeft + tab.offsetWidth;
+    const containerScrollRight = container.scrollLeft + container.offsetWidth;
+
+    // is tab behind (left)
+    if (container.scrollLeft > tab.offsetLeft) {
+      container.scrollLeft = tab.offsetLeft;
```
juanmitriatti, line 35: Hi @scott-williams-az 
Just asking, it sets container.scrollLeft directly, causing an instant jump.  For a better UX , should not we consider using container.scrollTo({ left: ..., behavior: 'smooth' }) (in case that's possible here).

juanmitriatti:
@scott-williams-az  PR looks good to me. I just left a minor comment.

packages/unity-bootstrap-theme/src/js/tabbed-panels.js
scott-williams-az: @juanmitriatti Your changes makes sense but I am wondering why it is an instant jump for you and I am seeing smooth already.

juanmitriatti: Good catch, the reason it looks smooth on your end is because .nav-tabs already has scroll-behavior: smooth in _tabbed-panels.scss (line 28). That CSS property makes any programmatic scrollLeft assignment animate automatically. My suggestion for scrollTo({ behavior: 'smooth' }) would just make the intent explicit in the JS rather than relying on the CSS implicitly, but since the CSS is already handling it, this is fine as-is.

davidornelas11 (approved):
Works as expected!

https://github.com/user-attachments/assets/138b6f1a-0180-4034-8b69-2caf3979faf8

davidornelas11:
WE do have an unused variable `LG_BREAKPOINT` leftover so  if you want to remove that, that would clean up some old code

juanmitriatti (approved):

PR #1687, opened 2026-04-28
UDS-1772: fix(component-header-footer): fix nav link doesn't close menu

scott-williams-az:
### Description

JSM - UDS - component-header nav link doesn't close (React, Other)
Single page apps do not reload page. Because href links leave the page, the menu is reset.
onClick links will now close the menu. 
Examples added to menu with onClick attributes instead of href.
Top level link should close (Mobile view) Menu
Dropdown level link should close (Desktop view) dropdown menu and (Mobile view) menu

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1772)


davidornelas11 (approved):
Created a quick story to test out the behavior and it works and menu closes as ecpected now when onclick is passed 

Image

PR #1708, opened 2026-05-22
fix(anchor-menu): use native button for the mobile toggle (WCAG 2.1.1…

juanmitriatti:
### Description

The mobile anchor menu toggle was an `<h2>` element with a `data-bs-toggle="collapse"` click handler. Two WCAG failures:

- **2.1.1 Keyboard (Level A)** — `<h2>` is not focusable, so keyboard-only users cannot open/close the menu.
- **4.1.2 Name, Role, Value (Level A)** — assistive technologies do not expose `<h2>` as a button, so the role and state are wrong.

**Solution:** wrap the toggle in a real `<button type="button">` nested inside the `<h2>` (W3C Disclosure pattern — heading outside, button inside). The button provides correct semantics and native keyboard support while the heading is preserved for screen-reader navigation. Styles for `.mobile-menu-toggler` reset the native button chrome (background, border, padding) and use `display: flex; justify-content: space-between` so the layout stays visually identical to before.

Applied in both paths:

- **Vanilla** (`unity-bootstrap-theme`): SCSS + story example + MDX usage docs.
- **React** (`unity-react-core`): JSX refactor + styled-component update; `aria-expanded` now mirrors `state.showMenu`.

**Testing steps**

1. Open the Anchor Menu story in Storybook at mobile width (≤ 991px).
2. Tab to the "On This Page:" toggle — focus ring must be visible.
3. Press `Enter` or `Space` — menu expands; `aria-expanded` flips to `true`; chevron rotates 180°.
4. Confirm the toggle looks identical to before (no native button background/border).
5. Resize to desktop (≥ 992px) — confirm no layout regression and chevron stays hidden.

## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2194)

### DIFF
--- a/packages/unity-bootstrap-theme/src/scss/extends/_anchor-menu.scss
+++ b/packages/unity-bootstrap-theme/src/scss/extends/_anchor-menu.scss
@@ -15,6 +15,28 @@
     top: 0;
   }
 
+  // Native <button> nested inside the heading; gives the toggle real
+  // keyboard and assistive-tech semantics (WCAG 2.1.1, 4.1.2) while
+  // keeping the heading available for screen-reader navigation.
+  .mobile-menu-toggler {
+    background: transparent;
+    border: 0;
+    padding: 0;
+    width: 100%;
+    text-align: inherit;
+    cursor: pointer;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+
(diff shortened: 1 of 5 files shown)


davidornelas11 (approved):
Tests out and fixes the wcag issues as expected

PR #1710, opened 2026-05-29
Uds 2197: update or remove pkg to support react 19 

scott-williams-az:
### Description
remove reactstrap
 - package is no longer updated and does not support react19
 - BreadcrumbItem converted to native LI element

update react-share
 - update prop quote to title
 - add aria-label
 - update email share to proper usage
   (drafts share email, rather than using mailto:authorEmail link)
 - add console warning if required article prop is undefined, but still allows render

## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2197)

### DIFF
--- a/packages/unity-react-core/src/components/Article/Article.jsx
+++ b/packages/unity-react-core/src/components/Article/Article.jsx
@@ -27,7 +27,7 @@ import { Wrapper, EventInfoWrapper } from "./Article.styles";
  */
 export const Article = ({
   type = "news",
-  articleUrl,
+  articleUrl = "",
   publicationDate,
   title,
   body,
@@ -182,9 +182,7 @@ export const Article = ({
           Location:
         </h4>
         {}
-        <div
-          dangerouslySetInnerHTML={sanitizeDangerousMarkup(`${eventLocation}`)}
-        />
+        <div dangerouslySetInnerHTML={sanitizeDangerousMarkup(eventLocation)} />
         {registrationUrl && zoomUrl && <a href={zoomUrl}>Attend on Zoom</a>}
       </div>
     );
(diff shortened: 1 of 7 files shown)


mlsamuelson (dismissed):
Approving. Thanks for this cleanup! 

Only thing I noticed is that the focus on the share icons has a rounded corner. As Article isn't used anywhere, not going to flag it as blocking. Marking approved.

packages/unity-react-core/src/components/Article/Article.jsx
```diff
@@ -168,27 +182,40 @@ export const Article = ({
...
+      </li>
     );
   };
-
+  if (!articleUrl) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      "Warning: articleUrl prop is required for Article component for social media sharing to work properly."
+    );
+    articleUrl = articleUrl || "";
```
davidornelas11: You can remopve the falsy check since you already checked above and this will always evaluate to the right operator. Also can be moved right after destructuring up top of the component

packages/unity-react-core/src/components/Article/Article.jsx
```diff
@@ -119,34 +118,49 @@ export const Article = ({
         <div className="col col-lg-4 col-md-6 col-12">
           <h4>Share this event:</h4>
           <div className="article-social-media">
-            <FacebookShareButton url={articleUrl} quote={title}>
+            <FacebookShareButton
+              url={`${articleUrl}`}
```
davidornelas11: Since this would show up as undefined if its not available, we should just do  `url={articleUrl}`

packages/unity-react-core/src/components/Article/Article.jsx
```diff
@@ -168,27 +182,40 @@ export const Article = ({
           Location:
         </h4>
         {}
-        <div dangerouslySetInnerHTML={sanitizeDangerousMarkup(eventLocation)} />
+        <div
+          dangerouslySetInnerHTML={sanitizeDangerousMarkup(`${eventLocation}`)}
```
davidornelas11: Same template literal undefined issue as above

davidornelas11 (changes requested):
Some small changes

davidornelas11 (approved):
looks good

PR #1712, opened 2026-06-04
fix(component-header-footer): fix arbitrary timeout in header submit …

davidornelas11:
…function

## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-1612)

### DIFF
--- a/packages/component-header-footer/src/header/components/UniversalNavbar/Search/index.js
+++ b/packages/component-header-footer/src/header/components/UniversalNavbar/Search/index.js
@@ -42,31 +42,31 @@ const Search = () => {
     /** @type {HTMLFormElement} */
     const form = e?.currentTarget;
     e.preventDefault();
-    /**
-     * Issue: Callback not currently available
-     * We need to ensure dataLayer events are being logged correctly
-     * Soluton might be in GA4 settings with a form event targeting the element ID
-     *
-     * This solution does not guarantee the event is logged before the page
-     * redirects.
-     * Preventing form submission with arbitrary timeout is always bad, but this
-     * may be small enough to not degrade the experience
-     *
-     * TODO: UDS-1612
-     */
+
+    let submitted = false;
+    const submit = () => {
+      if (!submitted && typeof form?.submit === "function") {
+        submitted = true;
+        form.submit();
+      }
+    };
+
     const searchInput =
       form && form.elements
         ? /** @type {HTMLInputElement|null} */ (form.elements.namedItem("q"))
         : null;
+
+    // Fallback: always submit within 2s regardless of GTM state. Useful for
+    // cases where GTM fails to load or execute for any reason, or if the user
+    // has blocked GTM.
+    setTimeout(submit, 2000);
+
     trackGAEvent({
       ...SEARCH_GA_EVENT,
       text: searchInput ? searchInput.value : "",
+      eventCallback: submit,
+      eventTimeout: 2000,
     });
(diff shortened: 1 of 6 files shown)


packages/shared/services/googleAnalytics.js
```diff
@@ -37,6 +39,8 @@ const trackGAEvent = ({
   text = "",
   region = "",
   component = "",
+  eventCallback,
+  eventTimeout,
```
scott-williams-az, line 43: Should eventTimeout have a numeric default = 0?

packages/shared/services/googleAnalytics.js
```diff
@@ -50,7 +54,13 @@ const trackGAEvent = ({
     text: text.toLowerCase(),
     component: component.toLowerCase(),
   };
-  if (dataLayer) dataLayer.push(e);
+  if (eventCallback) e.eventCallback = eventCallback;
```
scott-williams-az, line 57: const e, previously an object of strings, we are now sending a function to data layer (Just bringing this up for discussion).

packages/shared/services/googleAnalytics.js
```diff
@@ -50,7 +54,13 @@ const trackGAEvent = ({
...
+  if (eventTimeout) e.eventTimeout = eventTimeout;
+  if (dataLayer) {
+    dataLayer.push(e);
+  } else if (eventCallback) {
+    eventCallback();
```
scott-williams-az, line 62: We are adding a feature to this shared component, I don't think there is anything wrong with this, but I would like to discuss.

packages/component-header-footer/src/header/components/UniversalNavbar/Search/index.js
```diff
@@ -42,31 +42,31 @@ const Search = () => {
...
+
     const searchInput =
       form && form.elements
         ? /** @type {HTMLInputElement|null} */ (form.elements.namedItem("q"))
         : null;
+
+    // Fallback: always submit within 2s regardless of GTM state. Useful for
+    // cases where GTM fails to load or execute for any reason, or if the user
+    // has blocked GTM.
+    setTimeout(submit, 2000);
```
scott-williams-az, line 62: I see the comment about GTM state, however I think the function for trackGAEvent should always execute regardless of GTM ` setTimeout(submit, 2000);` not needed here.

packages/shared/services/googleAnalytics.js
scott-williams-az, line 62: `setTimeout(submit, 2000);`

Do we need to add a callback parameter and a timeout parameter?

Timeout could be added inside callback? Callback = ()=>{ setTimeout(submit, 2000); }

**GTM can see Form submission natively, without dataLayer method.**

**I know we have talked about this before, but I'd like to discuss again to make sure we have all options before adding a 2 second timeout to all submissions**

mlsamuelson (changes requested):
Merge conflict. Also questions from @scott-williams-az have not resolutions, so I'm unable to review with the knowledge these were discussed.

packages/component-header-footer/src/header/components/UniversalNavbar/Search/index.js
scott-williams-az: comment was added that addressed my concerns

scott-williams-az (approved):

mlsamuelson (approved):

PR #1719, opened 2026-06-16
feat(component-events): expose configurable noFeedText prop

juanmitriatti:
## Summary

Exposes the feed empty-state message as an optional, configurable `noFeedText` prop on the exported event card components. Previously it was hardcoded to `"No events to show."` inside `BaseFeed`, giving consumers (e.g. the Webspark/Drupal site) no way to customize it per placement or language.

Fully backward compatible — when `noFeedText` is omitted, the original `"No events to show."` text is rendered exactly as before.

## Changes

- `CardsGridEvents` / `CardsListEvents`: accept optional `noFeedText` (default `"No events to show."`) and forward it to `BaseFeed`.
- `BaseFeed`: passes `noFeedText` to `FeedContainerProvider` instead of the hardcoded string; added propType.
- `propTypes.js`: added `noFeedText: PropTypes.string` to `eventsPropTypes`.
- Storybook: added `WithCustomNoFeedText` story to both components (+ empty-feed mock handler).
- Tests: added coverage for default text and custom override on an empty feed.
- Bumped package to **3.2.0** (minor) and updated CHANGELOG.

## Passthrough for Drupal

`initCardsGridEventsComponent` / `initCardsListEventsComponent` already forward the entire `props` object, so consumers can supply `props.noFeedText` from `drupalSettings` with no further changes:

```js
Drupal.initCardsGridEventsComponent({
  targetSelector: '#my-events',
  props: { dataSource: { url: '...' }, noFeedText: 'No upcoming events.' }
});
```

## Testing

- `component-events` unit tests: **12 passed** (incl. 4 new `noFeedText` tests).
- Lint: 0 errors.
- Dist bundle rebuilt successfully.

> Note: the pre-push hook reported an unrelated failure in `component-header-footer` (a package not touched by this branch); this PR is scoped entirely to `component-events`.

### DIFF
--- a/packages/component-events/CHANGELOG.md
+++ b/packages/component-events/CHANGELOG.md
@@ -1,3 +1,10 @@
+# [@asu/component-events-v3.2.0](https://github.com/asu/asu-unity-stack/compare/@asu/component-events-v3.1.0...@asu/component-events-v3.2.0) (2026-06-16)
+
+
+### Features
+
+* **component-events:** expose configurable `noFeedText` prop on `CardsGridEvents` and `CardsListEvents` (defaults to "No events to show." for backward compatibility)
+
 # [@asu/component-events-v3.1.0](https://github.com/asu/asu-unity-stack/compare/@asu/component-events-v3.0.1...@asu/component-events-v3.1.0) (2026-04-03)
 
 
--- a/packages/component-events/CHANGELOG.md
+++ b/packages/component-events/CHANGELOG.md
@@ -1,10 +1,3 @@
-# [@asu/component-events-v3.2.0](https://github.com/asu/asu-unity-stack/compare/@asu/component-events-v3.1.0...@asu/component-events-v3.2.0) (2026-06-16)
-
-
-### Features
-
-* **component-events:** expose configurable `noFeedText` prop on `CardsGridEvents` and `CardsListEvents` (defaults to "No events to show." for backward compatibility)
-
 # [@asu/component-events-v3.1.0](https://github.com/asu/asu-unity-stack/compare/@asu/component-events-v3.0.1...@asu/component-events-v3.1.0) (2026-04-03)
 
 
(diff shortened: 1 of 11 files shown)


packages/component-events/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# [@asu/component-events-v3.2.0](https://github.com/asu/asu-unity-stack/compare/@asu/component-events-v3.1.0...@asu/component-events-v3.2.0) (2026-06-16)
+
+
+### Features
+
+* **component-events:** expose configurable `noFeedText` prop on `CardsGridEvents` and `CardsListEvents` (defaults to "No events to show." for backward compatibility)
+
```
mlsamuelson: These changes will be added by semantic release in the pipeline. Including them here may cause issues.

packages/component-events/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@asu/component-events",
-  "version": "3.1.0",
+  "version": "3.2.0",
```
mlsamuelson: This change should be added by semantic release in the pipeline.

mlsamuelson (dismissed):

mlsamuelson:
If an empty string is provided with the new prop, no message displays. I'm thinking a better behavior might be for it to fallback to the default string... but this may be intentional. Can you confirm?

packages/component-events/CHANGELOG.md
juanmitriatti: Reverted the version bump and changelog entry are removed; semantic-release will handle them as you mention.

juanmitriatti:
> If an empty string is provided with the new prop, no message displays. I'm thinking a better behavior might be for it to fallback to the default string... but this may be intentional. Can you confirm?

 Good catch , the empty-string behavior wasn't intentional. It was a side effect of the JS default parameter ("" bypasses it)

davidornelas11 (approved):
With michaels suggestion this looks good

PR #1726, opened 2026-06-22
Uds 2143 - Implement: MVP Accessibility Carousels

spaceperson:
## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2143)


packages/unity-react-core/src/components/ComponentCarousel/core/components/BaseCarousel/glide/glide.setup.js
```diff
@@ -293,20 +293,81 @@ function setupCaroarousel({
...
+        for (let j = 1; j < 9; j++) {
+          if (!header) {
+            header = slides[i].querySelector(`h${j}`);
+          }
+        }
+
+        let cardDiv = slides[i].querySelectorAll(`.card`)
+        if (cardDiv && cardDiv[0]) {
+          cardDiv[0].setAttribute("aria-live", "polite");
+          cardDiv[0].setAttribute("role", "alert");
```
davidornelas11, line 338: role="alert" plus an explicit aria-live="polite" on the same element conflicts (alert already implies assertive), and this announces the whole card body on every slide change instead of something short. Looks like #1763  already dropped this exact block for a cleaner aria-hidden/gliderTrack approach,  might be easiest to just pull that over instead of patching this in place.

packages/unity-react-core/src/components/ComponentCarousel/core/components/BaseCarousel/glide/glide.setup.js
```diff
@@ -293,20 +293,81 @@ function setupCaroarousel({
...
+        }
+
+
+        if (header) {
+          let allTexts = []
+          getText(slides[i], allTexts)
+          gliderElement.setAttribute("aria-labelledby", allTexts[0]);
```
davidornelas11, line 345: aria-labelledby needs to reference an element id, not raw text - allTexts[0] is just the heading's text content, so this won't resolve to anything for a screen reader. Also getText is called on the whole slide instead of just header, so it could grab text that comes before the heading.

packages/unity-react-core/src/components/ComponentCarousel/core/components/BaseCarousel/glide/glide.setup.js
```diff
@@ -293,20 +293,81 @@ function setupCaroarousel({
...
+          gliderElement.removeAttribute("aria-label");
+        } else {
+          gliderElement.setAttribute("aria-label", `Card ${i + 1}`);
```
davidornelas11, line 348: This fallback runs for any carousel without a heading in the slide, not just cards, an image/gallery slide with no caption would get announced as "Card 1". Might be worth making this generic ("Slide 1") instead.

packages/unity-react-core/src/components/ComponentCarousel/core/components/BaseCarousel/glide/glide.setup.js
```diff
@@ -293,20 +293,81 @@ function setupCaroarousel({
...
+        }
+
+        // slides[i].focus();
+
+      }
+    }
+
+
+    // Update bullet accessibility
+    const bullets = gliderElement.querySelectorAll(".glide__bullet");
```
davidornelas11, line 359: Heads up, for the image gallery `bullets .glide__bullet` matches the ``, not the `<button role="option">` wrapping it, so disabled/aria-current won't actually do anything there.

packages/unity-react-core/src/components/ComponentCarousel/core/components/BaseCarousel/index.jsx
```diff
@@ -79,14 +79,18 @@ const BaseCarousel = ({
 
   return (
     <div
-      role={role}
       aria-labelledby={ariaLabelledBy}
+      // role={role}
+      // role="group"
+      role="region"
```
davidornelas11, line 85: role is still a prop here but it's dead now that role is hardcoded to "region" - ImageGalleryCarousel passes role="figure" and that'll get silently overridden. Also the commented-out lines above can probably go.

packages/unity-react-core/src/components/ComponentCarousel/core/components/BaseCarousel/index.jsx
```diff
@@ -79,14 +79,18 @@ const BaseCarousel = ({
...
+      aria-label="Carousel pagination"
```
davidornelas11, line 86: This aria-label should be on the bullets group container, not here - that's the exact gap Andrew flagged on UDS-2150 (pagination group missing its label). BaseBulletItemContainer still only has role="group" with nothing on it.

davidornelas11 (commented):
Left a few inline comments. Main ones: aria-labelledby isn't pointing to an actual element id, the pagination group still doesn't have its aria-label (this was already flagged on UDS-2150), and role="figure" on the image gallery carousel gets silently dropped now. Also flagged the aria-live/role="alert" bit since #1763 already replaces it with a cleaner approach. Prev/next button disabled state work looks great though. One more thing worth a look (outside this diff): ImageBulletItems in BulletItems.jsx still has the old "Slide view X" label and aria-selected never updates - probably needs the same treatment at some point.

davidornelas11:
closed for #1774

PR #1736, opened 2026-07-01
Uds 2218 degree search links

oatkar:
### Description

 Update degree listing component so degree links point to degree search degree pages.

## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2218?atlOrigin=eyJpIjoiN2NmOThkZDUwY2JiNGFhODhmZmViM2EyN2I2MzE3M2MiLCJwIjoiaiJ9)


packages/app-degree-pages/src/core/services/degree-data-prop-resolver-service.jsx
```diff
@@ -85,6 +85,24 @@ function degreeDataPropResolverService(row = {}) {
...
+        UG:   "bachelors",
+        GR:   "masters-phd",
+        UGCM: "minors",
+      };
+
+      const programTypePath = programTypeMap[degreeType];
+
+      if (!programTypePath) return "";
+
+      return `https://degrees.apps.asu.edu/${programTypePath}/major/ASU00/${planCode}`;
```
mlsamuelson: In `packages/app-degree-pages/src/core/constants/web-api-constants.js` there is a DOMAIN_URL constant with the `https://degrees.apps.asu.edu` value defined (note, it doesn't provide the trailing slash). It would be good if we could use that constant here.

mlsamuelson (changes requested):
One update requested. After that, this looks good for the degree listing update.

packages/app-degree-pages/src/core/services/degree-data-prop-resolver-service.jsx
```diff
@@ -85,6 +85,24 @@ function degreeDataPropResolverService(row = {}) {
...
+
+      const programTypeMap = {
+        UG:   "bachelors",
+        GR:   "masters-phd",
+        UGCM: "minors",
+      };
+
+      const programTypePath = programTypeMap[degreeType];
+
+      if (!programTypePath) return "";
```
davidornelas11, line 103: This might be a question for @mlsamuelson or Andrew Nelson but what should happen for edge cases? what url do we want to fall back on as default or should we just not render a link and just make it plain text for those edge cases?

mlsamuelson: No link would be my vote. Don't think this should happen, but always good to have the logic.

davidornelas11: @oatkar Next steps would be to render plain text in this instance instead of a link with no href

mlsamuelson (approved):

PR #1739, opened 2026-07-03
fix(unity-bootstrap-theme): gold maroon gray and dark hover button br…

spaceperson:
…and standard color fix

## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2169)

### DIFF
--- a/packages/unity-bootstrap-theme/src/scss/extends/_buttons.scss
+++ b/packages/unity-bootstrap-theme/src/scss/extends/_buttons.scss
@@ -182,6 +182,50 @@ $button-colors: (
   }
 }
 
+.btn-gold:hover {
+  color: #000 !important;
+  background-color: #FFC627 !important;
+  border-color: #FFC627 !important;;
+}
+.btn-filled-gold:hover {
+  color: #000 !important;
+  background-color: #FFC627 !important;
+  border-color: #FFC627 !important;;
+}
+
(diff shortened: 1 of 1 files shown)


packages/unity-bootstrap-theme/src/scss/extends/_buttons.scss
```diff
@@ -182,6 +182,50 @@ $button-colors: (
   }
 }
 
+.btn-gold:hover {
```
davidornelas11, line 185: This approach works but i would suggest combining classnames if they ahve the same styles. example:
```scss
 .btn-gold:hover,
.btn-filled-gold:hover
    {
color: #000 !important;
  background-color: #FFC627 !important;
  border-color: #FFC627 !important;;
}
```
results in half the code needed

davidornelas11 (changes requested):
small suggestion for less code usage

packages/unity-bootstrap-theme/src/scss/extends/_buttons.scss
spaceperson: Something else in a CSS package has priority when combining the classnames.
Having them separate and set to !important is the only way to reliably overwrite CSS from that package to get the styles needed.

davidornelas11 (approved):

PR #1750, opened 2026-07-21
chore(static-site): update package

scott-williams-az:
### Description

just an update to fix build issues update static-site package to latest react-core


scott-williams-az:
This can be canceled, if another PR with the update is merged first.

scott-williams-az:
no longer needed

PR #1753, opened 2026-07-23
feat: modal accessibility and escape key features for unity react and…

spaceperson:
… bootstrap

## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2211)

### DIFF
--- a/packages/unity-react-core/src/components/Modal/Modal.tsx
+++ b/packages/unity-react-core/src/components/Modal/Modal.tsx
@@ -36,13 +36,66 @@ export const Modal: React.FC<ModalProps> = ({ open, gaData }) => {
   const { isReact, isBootstrap } = useBaseSpecificFramework();
   const [openState, setOpen] = React.useState(open);
 
+  const focusOnModalInput = () => {
+    //source: https://stackoverflow.com/questions/4195616/how-to-set-the-focus-on-a-javascript-modal-window
+    const focusableElements =
+      'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
+    const modal = document.getElementsByClassName("uds-modal-container")[0];
+    const firstFocusableElement = modal?.querySelectorAll(focusableElements)[0];
+    const focusableContent = modal?.querySelectorAll(focusableElements);
+    const lastFocusableElement = focusableContent
+      ? focusableContent[focusableContent?.length - 1]
+      : undefined;
+
+    if (lastFocusableElement && firstFocusableElement) {
+      document.addEventListener("keydown", function (e) {
+        let isTabPressed = e.key === "Tab" || e.keyCode === 9;
+
+        if (!isTabPressed) {
+          return;
+        }
+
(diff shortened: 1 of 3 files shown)


packages/unity-react-core/src/components/Modal/Modal.tsx
```diff
@@ -60,20 +113,23 @@ export const Modal: React.FC<ModalProps> = ({ open, gaData }) => {
       {(openState || isBootstrap) && (
         <div
           id="uds-modal"
+          role="dialog"
+          aria-modal="true"
+          aria-labelledby={modalTitle}
```
davidornelas11: `aria-labelledby` needs to reference an id this just points to a string of text. I would actually switch this to use `aria-label` instead of `aria-labelledby`.

packages/unity-react-core/src/components/Modal/Modal.tsx
```diff
@@ -36,21 +36,74 @@ export const Modal: React.FC<ModalProps> = ({ open, gaData }) => {
...
+          }
+        }
+      });
+      (firstFocusableElement as HTMLElement)?.focus();
+    }
+  };
+
   const handleOpen = () => {
     setOpen(true);
+    focusOnModalInput();
```
davidornelas11: `focusOnModalInput()` runs synchronously right after `setOpen(true)`, but React hasn't re-rendered yet. `#uds-modal-container` doesn't exist in the DOM on first open, so modal here resolves to undefined and no focus/trap gets set. Move this logic into a useEffect keyed on openState instead, so it runs after the DOM updates.

Claude suggested this refactor but feel free to come up with another solution if you dont think it works well:

```JSX
const handleOpen = () => {
  setOpen(true);
};

const handleClose = () => {
  setOpen(false);
};

useEffect(() => {
  if (!openState) return;

  const focusableElements =
    'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
  const modal = document.getElementsByClassName("uds-modal-container")[0];
  const focusableContent = modal?.querySelectorAll<HTMLElement>(focusableElements);
  const firstFocusableElement = focusableContent?.[0];
  const lastFocusableElement = focusableContent?.[focusableContent.length - 1];

  firstFocusableElement?.focus();

  const handleTabKey = (e: KeyboardEvent) => {
    if (e.key !== "Tab" || !firstFocusableElement || !lastFocusableElement) return;

    if (e.shiftKey && document.activeElement === firstFocusableElement) {
      lastFocusableElement.focus();
      e.preventDefault();
    } else if (!e.shiftKey && document.activeElement === lastFocusableElement) {
      firstFocusableElement.focus();
      e.preventDefault();
    }
  };

  document.addEventListener("keydown", handleTabKey);
  return () => document.removeEventListener("keydown", handleTabKey);
}, [openState]);
```

davidornelas11 (changes requested):
changes in comments

davidornelas11 (approved):
Approved

PR #1759, opened 2026-07-28
Webspark 2.19.0 hotfix - fix webdir nav controls - PROXY API

scott-williams-az:
This PR fixes the nav controls styling issue by reintroducing the styles into their own styled components files for unity-react-core and issues a new verison for app-webdir-ui and unity-react-core. Tests were also added for app0-webdir-ui as there was a fix added for anon images not showing up. 

Mock api data was also added for app-webdir-ui as the search endpoints no longer allow certain domains to query them.

Also:
- Split getBootstrapHTML (react-dom/server, Storybook/dev-tooling only)
  out of useBaseSpecificFramework.js into its own file. Every component
  imports useBaseSpecificFramework at runtime, so bundling
  react-dom/server there was pulling server-rendering internals
  (MessageChannel, TextEncoder, etc.) into every consumer's published
  dist for no runtime benefit, and broke jsdom-based tests in
  app-webdir-ui.

## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2240)
- [Similar PR with MOCK data](https://github.com/ASU/asu-unity-stack/pull/1759)

## Using a local proxy server, we can use real data
 - environment variables are used so local proxy is only used with local development
 - CORS is enabled for static-stire "https://asu.github.io/asu-unity-stack/"
 - future item can we enable CORS for "https://unity-uds-staging.s3.us-west-2.amazonaws.com"
   - staging site will not load the API until we fix that setting - Or we just need another way to allow testing interactions (sort/ filter / search)
image


davidornelas11 (approved):
Approved

PR #1762, opened 2026-07-31
feat(component-header-footer): add isUnbranded and unbrandedLogo props for unbranded (KE) sites

juanmitriatti:
## Summary

Adds two new optional props to the Global Header (`component-header-footer`), requested by Webspark for the Unbranded (KE) profile work (WS2-2964, WS2-3128). Jira: UDS-2248.

- `isUnbranded` (bool, default `false`)
- `unbrandedLogo` (object: `src`, `alt`, `brandLink`)

When `isUnbranded` is true:

- The ASU logo is replaced by `unbrandedLogo`, mirroring how `partnerLogo` is injected for `isPartner`.
- The universal navbar keeps only the skip-nav link and the report-accessibility link. The gray universal bar is not displayed unless one of those hidden links receives keyboard focus (focus-revealed via `:focus-within`, standard skip-link pattern).
- Sign In / Sign Out, ASU Home, My ASU, Colleges and Schools and the ASU search input are not rendered.
- Everything else (navTree menu, CTA buttons, title, parentOrg) behaves exactly as today.

Rationale: the universal nav links were hardcoded inside the React bundle with no prop to suppress them, so unbranded rendering was impossible without this change.

## Changes

- `ASUHeader`: new `isUnbranded` / `unbrandedLogo` props, wired through the app context, PropTypes, JSDoc typedefs and `types/main.d.ts`.
- `Logo`: renders `unbrandedLogo` (src/alt/brandLink) instead of the ASU logo when unbranded.
- `UniversalNavbar`: filters the nav links down to the two accessibility links, drops `Login` and `Search`, and adds a `uds-hdr-universal-nav-unbranded` class whose CSS visually hides the bar (still keyboard-reachable) except under `:focus-within`.
- `HeaderMain`: the mobile-menu search input is also suppressed when unbranded.
- New `Unbranded` Storybook story, updated `docs/README.props.md`.

## Testing

- 5 new unit tests in `header.test.js` (logo replacement, link reduction, sign-in/search suppression, focus-reveal class, nav/title regression). Full package suite: 51/51 passing, including the Playwright dataLayer tests.
- Verified manually in Storybook: gray bar hidden by default, revealed while tabbing through "Skip to main content" / "Report an accessibility problem", hidden again on blur.
- Note: local `yarn test` at the repo root shows a pre-existing, unrelated failure in `@asu/app-webdir-ui` (`MessageChannel is not defined` jsdom environment issue); this PR touches no files in that package.

### DIFF
--- a/packages/component-header-footer/src/header/components/UniversalNavbar/index.styles.js
+++ b/packages/component-header-footer/src/header/components/UniversalNavbar/index.styles.js
@@ -98,6 +98,22 @@ const Wrapper = styled.div`
       grid-template-columns: 1fr 1fr 1fr 1fr;
     }
   }
+
+  /* Unbranded (KE) sites: the gray bar stays visually hidden but keyboard
+  reachable, and is revealed only while a skip/accessibility link has focus. */
+  &.${CLASS_NAMES.UNIVERSAL_NAV_UNBRANDED}:not(:focus-within) {
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    min-height: 0;
+    padding: 0;
+    margin: -1px;
+    overflow: hidden;
+    clip: rect(0 0 0 0);
+    clip-path: inset(100%);
+    white-space: nowrap;
+    border: 0;
+  }
 `;
 
 export { Wrapper };
--- a/packages/component-header-footer/src/header/components/UniversalNavbar/index.styles.js
+++ b/packages/component-header-footer/src/header/components/UniversalNavbar/index.styles.js
@@ -99,7 +99,7 @@ const Wrapper = styled.div`
     }
   }
 
-  /* Unbranded (KE) sites: the gray bar stays visually hidden but keyboard
+  /* Unbranded sites: the gray bar stays visually hidden but keyboard
   reachable, and is revealed only while a skip/accessibility link has focus. */
   &.${CLASS_NAMES.UNIVERSAL_NAV_UNBRANDED}:not(:focus-within) {
     position: absolute;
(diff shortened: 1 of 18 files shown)


mlsamuelson (changes requested):
In the footer, if the ContactComponent renders (Webspark has an option to enable/disable it), the button in that component has a hardcoded "Support ASU" label.

Two possible solutions:
1. add a label prop for that button. Will also require WS2 side-changes.
2. remove the button if isUnbranded is true.

I'm leaning towards simplicity, and suggesting 2.

mlsamuelson (changes requested):
Do not display the footer colophon when unbranded - the gray bar with Copyright and accessibility and other links should not display. If needed, on an unbranded site, the site owner can add a block to add their own custom colophon.

mlsamuelson (changes requested):
In the unbranded stories, instead of "KE" and "Knowledge Enterprise" let's just use "Unbranded"

packages/component-header-footer/src/header/components/UniversalNavbar/index.styles.js
```diff
@@ -98,6 +98,22 @@ const Wrapper = styled.div`
       grid-template-columns: 1fr 1fr 1fr 1fr;
     }
   }
+
+  /* Unbranded (KE) sites: the gray bar stays visually hidden but keyboard
```
mlsamuelson: Let's not mention KE here and in the code. Unbranded is unbranded, doesn't matter who it's for. Eventually other teams will use it.

mlsamuelson (changes requested):

juanmitriatti:
Hi @mlsamuelson 
Thanks for the review!  Addressed all the feedback.

mlsamuelson (approved):

PR #1767, opened 2026-08-04
feat(component-header-footer): read brand colors from CSS custom properties (UDS-2250)

juanmitriatti:
### What & why
Webspark's Unbranded (KE) profile (epic WS2-2964) replaces brand colors in the compiled `unity-bootstrap-theme` stylesheet, so every SCSS-styled component follows the site palette. The header and footer did **not**: their colors were hardcoded as styled-components JS constants injected at runtime, out of reach of any stylesheet replacement.

This change makes `component-header-footer` read every **brand** color from a CSS custom property with the current value as fallback, e.g. `var(--bs-gold, #ffc627)`. Branded sites are **pixel-identical** because of the fallbacks; sites that define the variables (Webspark unbranded palettes rewrite `--bs-*` in the compiled CSS) get the header and footer following the palette natively.

Pairs with UDS-2248 (`isUnbranded`/`unbrandedLogo` props).

### Changes
- **`src/header/colors.js`** — the two brand constants now read from custom properties (single choke point; all 10 header style files and the `ACCENT_*` aliases inherit it, gradients included):
  - `ASU_MAROON` → `var(--bs-maroon, #8c1d40)`
  - `ASU_GOLD` → `var(--bs-gold, #ffc627)`
- **`src/footer/index.styles.js`** — the footer's local `--color-base-gold` definition now reads `var(--bs-gold, #ffc627)`; every `var(--color-base-gold)` usage (Innovation background, contact/accordion gold hovers) inherits it.
- **README.md** — new "Theming: brand colors" section documenting the pattern and rationale.
- Added tests locking the contract: `src/header/colors.test.js` and `src/footer/footer-brand-colors.test.js`.

### Scope decisions
- **Brand accents only** (maroon, gold). Grays/neutrals stay as fixed literals — the unbranded palette swaps brand colors, not neutrals.
- **No `unity-bootstrap-theme` change.** `--bs-maroon`/`--bs-gold` (and `--bs-darkmaroon`/`--bs-darkgold`) are already emitted at `:root` and already rewritten by the Webspark palette system; no header/footer style uses a dark shade, so no `--uds-color-base-*` additions were needed.
- **No new props, no API changes, no new dependencies.**

### Testing
- `component-header-footer` Jest suite: **50 passed / 10 suites** (includes the 2 new contract tests).
- Datalayer Playwright: **2 passed**.
- The `footer-brand-colors` test was mutation-verified (reverting the source to bare hex fails the test).

### Acceptance criteria notes
- [ ] Accessibility (keyboard + screen reader) — no rendered change on branded sites; palette-driven change is Webspark's responsibility.
- [ ] Design/Brand review — branded output is byte-identical, so no Brand input expected.
- [x] Technical documentation — README section added.
- [ ] Cross-team code review via PR.

### DIFF
--- a/packages/component-header-footer/src/header/colors.js
+++ b/packages/component-header-footer/src/header/colors.js
@@ -6,8 +6,11 @@
  */
 
 // Primary ASU Brand Colors
-export const ASU_MAROON = "#8c1d40";
-export const ASU_GOLD = "#ffc627";
+// Read from Bootstrap-emitted CSS custom properties so Webspark's unbranded (KE)
+// palette (which rewrites --bs-* in the compiled unity-bootstrap-theme CSS) flows
+// into the header. The hex fallback keeps branded sites pixel-identical. UDS-2250.
+export const ASU_MAROON = "var(--bs-maroon, #8c1d40)";
+export const ASU_GOLD = "var(--bs-gold, #ffc627)";
 
 // Grayscale Colors
 export const ASU_WHITE = "#ffffff";
--- a/packages/component-header-footer/src/header/colors.js
+++ b/packages/component-header-footer/src/header/colors.js
@@ -6,7 +6,7 @@
  */
 
 // Primary ASU Brand Colors
-// Read from Bootstrap-emitted CSS custom properties so Webspark's unbranded (KE)
+// Read from Bootstrap-emitted CSS custom properties so Webspark's unbranded
 // palette (which rewrites --bs-* in the compiled unity-bootstrap-theme CSS) flows
 // into the header. The hex fallback keeps branded sites pixel-identical. UDS-2250.
 export const ASU_MAROON = "var(--bs-maroon, #8c1d40)";
--- a/packages/component-header-footer/README.md
+++ b/packages/component-header-footer/README.md
@@ -41,7 +41,7 @@ usages such as the selected nav-item underline and the animated title underline.
 
 - **Branded sites** render pixel-identically: with no `--bs-*` overrides present,
   the hex fallback applies.
-- **Unbranded (KE) sites**: Webspark's unbranded palette rewrites `--bs-*` in the
+- **Unbranded sites**: Webspark's unbranded palette rewrites `--bs-*` in the
   compiled `unity-bootstrap-theme` stylesheet, so the header and footer follow the
   site palette natively — no props or API changes required.
 
--- a/packages/component-header-footer/README.md
+++ b/packages/component-header-footer/README.md
@@ -31,3 +31,20 @@ import { ASUHeader as Header, ASUFooter as Footer } from '@asu/component-header-
 
 ### Import for use in HTML page
 You can find an example of how to set `ASUHeader` and `ASUFooter` props for use in a browser [here](/packages/component-header/examples/global-header-footer.html)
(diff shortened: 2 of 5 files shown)


scott-williams-az (dismissed):

packages/component-header-footer/src/header/colors.js
```diff
@@ -6,8 +6,11 @@
  */
 
 // Primary ASU Brand Colors
-export const ASU_MAROON = "#8c1d40";
-export const ASU_GOLD = "#ffc627";
+// Read from Bootstrap-emitted CSS custom properties so Webspark's unbranded (KE)
```
mlsamuelson: Don't mention KE in Unbranded code.

packages/component-header-footer/src/header/colors.js
```diff
@@ -6,8 +6,11 @@
...
+// palette (which rewrites --bs-* in the compiled unity-bootstrap-theme CSS) flows
+// into the header. The hex fallback keeps branded sites pixel-identical. UDS-2250.
+export const ASU_MAROON = "var(--bs-maroon, #8c1d40)";
```
mlsamuelson, line 12: Are we adding a dependency on BS here when the header component is designed to be standalone (at runtime)? Or is this a compile-time action only? Should we avoid using BS in the variable name?

mlsamuelson (changes requested):

packages/component-header-footer/src/header/colors.js
juanmitriatti: On the --bs-* naming: you're right, and regarding the dependency question, it's runtime CSS custom-property resolution, not a hard dependency. var(--bs-gold, #ffc627) resolves in the browser, and if --bs-* isn't defined anywhere the hex fallback kicks in as you know.
Anyways,  I agree that naming the variable --bs-* couples a standalone component to Bootstrap's namespace, which isn't great.

I reached for --bs-* only because Webspark's unbranded palette already rewrites those in the compiled theme, so it "just works" with zero Webspark-side changes.
What do you think about a layered fallback that gives us a clean, own namespace and keeps that automatic pickup?

```
export const ASU_MAROON = "var(--uds-header-maroon, var(--bs-maroon, #8c1d40))";
export const ASU_GOLD   = "var(--uds-header-gold, var(--bs-gold, #ffc627))";
```

This way, we have our own naming convention, while Webspark can keep working as it does today without any changes. Also, the fallback also makes sure the current colors stay the same. I'll wait for your confirmation @mlsamuelson

mlsamuelson: My main concern is addressed - that we're not introducing a hard dependency. The --bs-* naming I'm less concerned about since we're going to be working on the Lit transition and removing more of the Bootstrap stuff here soon, so I'd be okay with it staying as it is and we can address the namespacing in that larger effort, as things are likely to change some with that anyway.

mlsamuelson (approved):

packages/component-header-footer/src/header/colors.js
juanmitriatti: Understood.

juanmitriatti:
This PR is ready to merge

PR #1768, opened 2026-08-05
feat(unity-react-core): modal auto focus and disabled tab to content …

spaceperson:
…outside the modal

## Checklist

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2211)


packages/unity-react-core/src/components/Modal/Modal.tsx
```diff
@@ -144,7 +160,60 @@ export const Modal: React.FC<ModalProps> = ({ open, gaData }) => {
...
+                  id="closeModalButtonR"
+                  onClick={handleClose}
+                  // data-bs-dismiss={isBootstrap && "modal"}
+                  className="uds-modal-close-btn"
+                  icon={["fas", "times"]}
+                />
+              </GaEventWrapper>
+              <h1>{modalTitle}</h1>
+              <p>
+                Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
```
scott-williams-az, line 206: Content should be a prop

packages/unity-react-core/src/core/types/shared-types.js
```diff
@@ -24,6 +24,7 @@
  * @typedef {Object} ButtonIconOnlyProps
  * @property {Array.<string>} icon
  * @property {string} [color]
+ * @property {boolean} [autoFocus]
```
scott-williams-az, line 27: Should this always be true?

packages/unity-react-core/src/components/Modal/Modal.tsx
```diff
@@ -57,9 +53,19 @@ export const Modal: React.FC<ModalProps> = ({ open, gaData }) => {
   }, [openState, setOpen]);
 
   useEffect(() => {
-    if (!openState) return;
+    if (!openState) {
```
scott-williams-az, line 56: @davidornelas11 are we mixing our react components with js inside bootstrap-theme (it technically loses react "coolness" but makes the functionality consistant)

If we use react functionality,
- make sure the bootstrap JS script does not conflict.
- useRef will be a better solution to target the element.

packages/unity-bootstrap-theme/src/js/modals.js
```diff
@@ -1,22 +1,101 @@
 import { EventHandler } from "./bootstrap-helper";
 
+function openModal() {
```
scott-williams-az, line 3: Should the modal close if a user clicks the backdrop? I thought this was a standard behavior?

packages/unity-bootstrap-theme/src/js/modals.js
```diff
@@ -1,22 +1,101 @@
...
+      closeModalButton?.focus();
+    }
+  }, 200);
+
+  // const mainContent = document.getElementById("main-content");
+  // mainContent.setAttribute("inert", "");
+  // const mainContentChildren = mainContent.children;
+  // let mainContentChildrenArray = Array.from(mainContentChildren);
+  // for (let i = 0; i < mainContentChildrenArray.length; i++) {
+  //   mainContentChildrenArray[i].setAttribute("inert", "");
```
scott-williams-az, line 18: clean up code we aren't keeping

packages/unity-bootstrap-theme/src/js/modals.js
```diff
@@ -1,22 +1,101 @@
...
     });
 
   document
     .getElementById("closeModalButton")
     ?.addEventListener("click", function () {
-      document.getElementById("uds-modal").classList.remove("open");
+      closeModal();
     });
 
   document?.addEventListener("keydown", function (event) {
```
scott-williams-az, line 95: remove this event when the modal is not open

packages/unity-react-core/src/core/types/shared-types.js
scott-williams-az: Oh the name confused me. is this variable for automatically opening the modal on page load? Can we pick a different variable name? variable name should describe if true the modal will be open when it loads

packages/unity-bootstrap-theme/src/js/modals.js
```diff
@@ -1,22 +1,101 @@
...
+
+  const handleTabKey = e => {
+    let isTabPressed = e.key === "Tab"; // || e.keyCode === 9;
+
+    if (!isTabPressed) {
+      return;
+    }
+
+    if (e.shiftKey) {
+      // if shift key pressed for shift + tab combination
```
scott-williams-az, line 40: lets let the native tab event do its thing. you can create a focus trap


```
import { EventHandler } from "./bootstrap-helper";

function initModals() {
  const modal = document.getElementById("uds-modal");
  const openModalButton = document.getElementById("openModalButton");
  const closeModalButton = document.getElementById("closeModalButton");
  const firstFocusable = modal.querySelector(
    'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'
  );

  let previousFocus = null;
  function focusTrap(e) {
    // If relatedTarget is outside, move focus back inside
    if (!modal.contains(e.relatedTarget)) {
      firstFocusable?.focus();
    }
  }

  function openModal() {
    // When opening: save current focus and move into modal
    previousFocus = document.activeElement;
    modal.classList.add("open");
    // attach event listeners to trap focus and close modal
    modal.addEventListener("focusout", focusTrap);
    modal.addEventListener("click", closeModal, true);
    document.addEventListener("keydown", closeModal);
    // Focus the first interactive element inside (or body if none)
    firstFocusable?.focus();
  }

  function closeModal({ type, target, key } = {}) {
    if (
      // escape key pressed
      (type === "keydown" && key === "Escape") ||
      // click outside modal
      (type === "click" && target === modal) ||
      // click on close button
      (type === "click" && target === closeModalButton)
    ) {
      modal.classList.remove("open");
      modal.removeEventListener("focusout", focusTrap);
      modal.removeEventListener("click", closeModal, true);
      document.removeEventListener("keydown", closeModal);
      // When closing: restore original focus
      previousFocus?.focus();
    }
  }
  openModalButton.addEventListener("click", openModal);
}

EventHandler.on(window, "load.uds.modals", initModals);

export { initModals };
```

packages/unity-react-core/src/core/types/shared-types.js
spaceperson: It's an html value for automatically updating the focus to activate the button when the modal opens. I left it as an option, but it didn't work when I tried it. I assumed it was because another element was taking priority. I didn't see the harm in leaving it as an option, but we could remove it.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus
JSX uses camel case instead of the lowercase autofocus value.

packages/unity-bootstrap-theme/src/js/modals.js
spaceperson: I thought that feature missing was odd as well but wasn't sure if there was a reason we didn't have that feature. I'll implement that with the next update.

PR #1783, opened 2026-09-11
fix(component-header-footer): uds-2264 - search input padding

chill5-0:
### Padding-left for the search input box should be .5rem on both desktop and mobile.

To reproduce - Type some text into the top right universal search field (gray bar) and view the padding-left via browser inspector - should be .5rem, or 8px.

### Links

- [JIRA ticket](https://asudev.jira.com/browse/UDS-2264)


davidornelas11 (approved):
Tested in webspark and works as expected

